PR #3, opened 2020-05-31
Feature/firestore

evilpenguin:


Say Their Names/Source/Log/Log.swift
```diff
@@ -0,0 +1,74 @@
+//
+//  Log.swift
+//  Say Their Names
+//
+//  Created by evilpenguin on 5/31/20.
+//  Copyright © 2020 Franck-Stephane Ndame Mpouli. All rights reserved.
+//
+
+import UIKit
```
Kilo-Loco, line 9: Delete duplicate please

Say Their Names/Source/Frameworks/ServiceFrameworks.swift
```diff
@@ -0,0 +1,20 @@
...
+//  Copyright © 2020 Franck-Stephane Ndame Mpouli. All rights reserved.
+//
+
+import UIKit
+import Firebase
+
+// MARK: - ServiceFrameworks
+
+typealias ServiceFrameworks = Service
+extension ServiceFrameworks {
```
Kilo-Loco, line 15: Whats the benefit of using a `typealias` here?

evilpenguin: There is none outside of abstraction, and not just have `extension class` everywhere.

Say Their Names/Source/Log/Log.swift
evilpenguin: Removed

Say Their Names/Source/Frameworks/ServiceFrameworks.swift
Kilo-Loco: If you think this makes things easier from a readability pov then let's go ahead and keep it. I'll leave it up to you

evilpenguin: We can always remove it later. Simple refactor in the feature.

PR #9, opened 2020-06-01
Added support for service dependency injection

evilpenguin:


Say Their Names/Source/Components/MainTabBarController.swift
```diff
@@ -9,7 +9,8 @@
 import UIKit
 
 class MainTabBarController: UITabBarController, UITabBarControllerDelegate {
-
+    public weak var service: Service!
```
Kilo-Loco, line 12: We should remove the `!` from `Service` so we dont have to deal with optionals

Say Their Names/Source/Controller/Base/BaseViewController.swift
```diff
@@ -0,0 +1,20 @@
...
+//  Say Their Names
+//
+//  Created by evilpenguin on 5/31/20.
+//  Copyright © 2020 Franck-Stephane Ndame Mpouli. All rights reserved.
+//
+
+import UIKit
+
+class BaseViewController: UIViewController {
+    public weak var service: Service!
```
Kilo-Loco, line 12: We can add a new coder initializer that will accept the `Service` as a parameter so we dont need a force unwrapped `Service`
```
init?(coder: NSCoder, service: Service) {
    self.service = service
    super.init(coder: coder)
}
```

Say Their Names/Source/Components/MainTabBarController.swift
evilpenguin: The variable is weakly reference. Removing the `!` would require it to be `?` or remove the `weak` qualifier.

Kilo-Loco: Oh thats the other thing. Do we actually need to have a weak reference to the service? The reference count should be decremented automatically when the Controller is deallocated, no?

evilpenguin: Indeed. As long as we keep `Service` to not contain any references to a vc... then using a strong reference is acceptable.

Kilo-Loco: Yea, I really dont think the Service should reference a VC so I think it might be a good move to make it a strong reference

evilpenguin: Removed.

Say Their Names/Source/Controller/Base/BaseViewController.swift
evilpenguin: Agree, but I didn't go the coder approach. Pushed changes.

PR #16, opened 2020-06-01
Load and parse people and donations from Firestore

hybridcattt:
Super basic code that allows to get the main list, and for each person get an array of donations. 

The idea is that donations would be fetched on each individual users screen when it's opened. 

Still experimenting with data structure of firebase so this is not final, but feel free to modify or try using it!

### DIFF
--- a/Say Their Names/Source/Frameworks/FirebaseIntegration.swift
+++ b/Say Their Names/Source/Frameworks/FirebaseIntegration.swift
@@ -23,7 +23,6 @@ final class FirebaseIntegration {
         if isFirebaseConfigured == false {
             FirebaseApp.configure()
             isFirebaseConfigured = true
-            
         }
         return Firestore.firestore()
     }
@@ -34,20 +33,45 @@ final class FirebaseIntegration {
     // https://firebase.google.com/docs/firestore/query-data/get-data
     
     func getPeople(completion: @escaping (Result<[Person], Error>) -> Void) {
-        db.collection("people").getDocuments { (snapshot, error) in
+        let peopleRef = db.collection("people")
+        
+        peopleRef.fetch { (result: Result<[Person], Error>) in
+            completion(result)
+        }
+    }
+    
+    func getDonations(person: Person, completion: @escaping (Result<[Donation], Error>) -> Void) {
+        let donationIds = person.donations.map( { $0.documentID } )
+        let relatedDonationsRef = db.collection("donation").whereField(FieldPath.documentID(), in: donationIds)
+        
+        relatedDonationsRef.fetch { (result: Result<[Donation], Error>) in
+            completion(result)
+        }
+    }
+}
+
+
+// MARK: - Reusable -
+
+fileprivate extension Query {
+    
+    func fetch<T: Codable>(_ type: T.Type = T.self, completion: @escaping (Result<[T], Error>) -> Void) {
+        
+        self.getDocuments { (snapshot: QuerySnapshot?, error) in
             if let error = error {
                 completion(.failure(error))
             } else if let snapshot = snapshot {
-                let people = snapshot.documents.compactMap({ (documentSnapshot: QueryDocumentSnapshot) -> Person? in
+                let objects = snapshot.documents.compactMap({ (documentSnapshot: QueryDocumentSnapshot) -> T? in
+                    print("Parsing data: ", documentSnapshot.data())
                     do {
-                        return try documentSnapshot.data(as: Person.self)
+                        return try documentSnapshot.data(as: T.self)
                     }
                     catch {
                         print(error)
                         return nil
                     }
                 })
-                completion(.success(people))
+                completion(.success(objects))
             }
         }
     }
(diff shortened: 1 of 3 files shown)


Say Their Names/Source/Frameworks/FirebaseIntegration.swift
```diff
@@ -34,20 +33,45 @@ final class FirebaseIntegration {
     // https://firebase.google.com/docs/firestore/query-data/get-data
     
     func getPeople(completion: @escaping (Result<[Person], Error>) -> Void) {
-        db.collection("people").getDocuments { (snapshot, error) in
+        let peopleRef = db.collection("people")
```
Kilo-Loco, line 36: If we go the route of Firebase, can we make the references instance properties of `FirebaseIntegration`?

Say Their Names/Source/Frameworks/FirebaseIntegration.swift
```diff
@@ -34,20 +33,45 @@ final class FirebaseIntegration {
...
+        
+        peopleRef.fetch { (result: Result<[Person], Error>) in
+            completion(result)
+        }
+    }
+    
+    func getDonations(person: Person, completion: @escaping (Result<[Donation], Error>) -> Void) {
+        let donationIds = person.donations.map( { $0.documentID } )
+        let relatedDonationsRef = db.collection("donation").whereField(FieldPath.documentID(), in: donationIds)
```
Kilo-Loco, line 45: Same here

Say Their Names/Source/Frameworks/FirebaseIntegration.swift
```diff
@@ -34,20 +33,45 @@ final class FirebaseIntegration {
...
+        }
+    }
+}
+
+
+// MARK: - Reusable -
+
+fileprivate extension Query {
+    
+    func fetch<T: Codable>(_ type: T.Type = T.self, completion: @escaping (Result<[T], Error>) -> Void) {
```
Kilo-Loco, line 58: Great approach here! 😁

Say Their Names/Source/Model/Person.swift
```diff
@@ -7,23 +7,30 @@
...
+    let dateOfBirth: Date
+
+    let location: DocumentReference
+    
+//    let media: [String]
+    let picture: String
+    let bio: String
+    let context: String
+    let donations: [DocumentReference]
+    let petitions: [DocumentReference]
```
Kilo-Loco, line 29: Haven't tried this approach before, but wouldn't it be better to work with first party objects as opposed to `DocumentReference`s ?

Say Their Names/Source/Model/Person.swift
```diff
@@ -7,23 +7,30 @@
...
     
     enum CodingKeys: String, CodingKey {
-        case id, fullName = "full_name", age, childrenCount = "children_count",
-        date, location, media, bio, context, donations, petitions
+        case /*id,*/ name , age, numberOfChildren = "number_of_children", dateOfBirth = "date_of_birth",
```
Kilo-Loco, line 32: When working with Firebase in the past, I found it useful to keep a reference to the object ID. Should we keep ID as part of the objects?

Say Their Names/Source/Frameworks/FirebaseIntegration.swift
hybridcattt: Sure, that’s a nice idea!

Say Their Names/Source/Model/Person.swift
hybridcattt: As far as I understand there is no “deep fetch” feature in Firestore. So we only get references and not full objects, needing to fetch them separately. 
I would prefer to have fetched objects directly too!
One option is to duplicate the data, but I was hesitant to start with that optimization right away.
What do you think?

Say Their Names/Source/Model/Person.swift
hybridcattt: Each object (document) gets a unique documentID, so we don’t have to have a separate field. Though we can have a custom field too if we want to. Do you think it can be useful?

Say Their Names/Source/Model/Person.swift
evilpenguin: We should use https://github.com/alickbass/CodableFirebase.

hybridcattt: We can try that as a next step from here 👍 so far I followed official documentation

Say Their Names/Source/Model/Person.swift
Kilo-Loco: I think it becomes easier to make related queries when the `id` is stored. For example, we could query against a different collection that has documents that keep a map of people and get back each of those documents that have the matching `personId`

Say Their Names/Source/Model/Person.swift
Kilo-Loco: IMO our first party objects should resemble the document's structure; meaning that we should only keep an array of `donationIds` and `petitionIds` if that is how we plan on structuring the data

hybridcattt:
Not pursuing the solution with Firebase, so this is irrelevant

PR #26, opened 2020-06-03
Refactor PetitionsController to use programmatic UI

Kilo-Loco:
This commit is being used to propose a potential MVC design pattern.

- ViewController is responsible for utilizing services, handling business logic, and providing the view with data
- View is responsible for laying out subviews and displaying the data that is provided to it. (Views should be dumb and should not implement any business logic)

**PLEASE COMMENT ON THIS PR IF YOU DISAGREE WITH THIS PROPOSAL**


jaywardell:
I am doing something very similar in https://github.com/Say-Their-Name/ios/tree/HomeController-programmatic-Layout.  Main difference is I broke out a protocol "ServiceClient" from BaseViewController.

kingsqd:
I like the approach. One question, do you think we should rename PetitionUI to be something more general so that it can be re-used by the dontations UI. they seem to be very similar according to the mocks

FranckNdame:
> I like the approach. One question, do you think we should rename PetitionUI to be something more general so that it can be re-used by the dontations UI. they seem to be very similar according to the mocks

Agreed.

Say Their Names/Source/Controller/Petitions/PetitionsController.swift
```diff
@@ -8,8 +8,21 @@
 
 import UIKit
 
-class PetitionsController: BaseViewController {
-    override func viewDidLoad() {
-        super.viewDidLoad()
+/// Controller responsible for showing the petitions
+final class PetitionsController: BaseViewController {
+    
+    private let ui: PetitionsUI
```
unnamedd, line 14: We should avoid abbreviations and raw variables. Here, the best name is `petitionsUI` or maybe `viewCode`.
Also, we might want to make it lazy var

Say Their Names/Source/Controller/Petitions/PetitionsUI.swift
```diff
@@ -0,0 +1,32 @@
...
+//  Say Their Names
+//
+//  Created by Kyle Lee on 6/2/20.
+//  Copyright © 2020 Franck-Stephane Ndame Mpouli. All rights reserved.
+//
+
+import UIKit
+
+/// The UI for Petitions
+final class PetitionsUI: UICollectionView {
```
unnamedd, line 12: To have some consistency with the type used, it will make more sense to name it as `PetitionCollectionView` instead of `PetitionsUI`, after all, this is only a CollectionView and does not reflect the whole content of the whole screen. For the viewCode of the whole screen, we should have something like `PetitionViewCode` way.

Say Their Names/Source/Controller/Petitions/PetitionsController.swift
hybridcattt: another idea is to have something  `private let contentView: PetitionsView` to be clear with the purpose of the variable

hybridcattt:
it's middle of the night at @Kilo-Loco 's location, @unnamedd  would you like to take over and implement the fixes?

unnamedd:
> it's middle of the night at @Kilo-Loco 's location, @unnamedd would you like to take over and implement the fixes?

Even if I can do it @hybridcattt, I guess the benefit of the code review is the sharing knowledge. I can do, but then I won't have his opinion about it. I rather wait for him.

hybridcattt:
Your suggestions are minor and very reasonable so I'm sure @Kilo-Loco would not disagree! This project is collaborative so we just try to do our best to move the project forward and release it as soon as reasonably possible 🙂

Say Their Names/Source/Controller/Petitions/PetitionsViewCode.swift
```diff
@@ -0,0 +1,65 @@
...
+    }
+    
+    required init?(coder: NSCoder) {
+        fatalError()
+    }
+    
+    override func layoutSubviews() {
+        super.layoutSubviews()
+        
+        setupSubviews()
```
hybridcattt, line 34: `layoutSubviews` is called quite often, and the setup needs to happen only once. So it's better to put it in `setupSelf` method instead

Say Their Names/Source/Controller/Petitions/PetitionsViewCode.swift
```diff
@@ -0,0 +1,65 @@
...
+//  Created by Kyle Lee on 6/2/20.
+//  Copyright © 2020 Franck-Stephane Ndame Mpouli. All rights reserved.
+//
+
+import UIKit
+
+/// The UI for Petitions
+final class PetitionsViewCode: UIView {
+    
+    private lazy var petitionsLabel: UILabel = {
```
hybridcattt, line 14:
```suggestion
    private let petitionsLabel: UILabel = {
```
will work the same, just without the ability to accidentally overwrite the variable 🙂

Say Their Names/Source/Controller/Petitions/PetitionsViewCode.swift
```diff
@@ -0,0 +1,65 @@
...
+        let label = UILabel()
+        label.text = "Petitions"
+        label.textAlignment = .center
+        
+        return label
+    }()
+    
+    init() {
```
hybridcattt, line 22: `init(frame: CGRect)` will work exactly the same, allowing to use both `init()` and `init(frame:)`. It's due to how initializers are inherited, can read more here (though it's a long read) : https://docs.swift.org/swift-book/LanguageGuide/Initialization.html

Say Their Names/Source/Controller/Petitions/PetitionsUI.swift
Kilo-Loco: As of right now, the designs call for only a collection view, but this may change over time. Maybe it would be best for me to change this to a view that has a collection view subview

Say Their Names/Source/Controller/Petitions/PetitionsController.swift
Kilo-Loco: I chose `ui` as it could be used as a universal member across all the different views we will have. I think the next best thing would be `petitionsView` as it is more clear.

PR #27, opened 2020-06-03
Implement PersonPhotoCell programmatically

leonardo-garcia:
## What does this PR do?
This PR removes the implementation of PersonPhotoCell made with XIBs and implements it programmatically.

I added a protocol called Reusable that I use to simplify the registration and dequeueing of cells, although if you do not agree with it I can remove it.

I also updated the XIB of PersonDetailsController to match the design found on Figma. Nonetheless this will get removed once it is implemented with code.

## Ticket 
https://trello.com/c/0xP30el2/26-refactor-personphotocell-to-use-programmatic-ui

## Screenshot

image


Say Their Names/Source/Controller/Home/Person/PersonDetailsController.swift
```diff
@@ -41,12 +40,13 @@ extension PersonDetailsController: UICollectionViewDataSource, UICollectionViewD
     }
     
     func collectionView(_ collectionView: UICollectionView, cellForItemAt indexPath: IndexPath) -> UICollectionViewCell {
-        let cell = collectionView.dequeueReusableCell(withReuseIdentifier: "PersonPhotoCell", for: indexPath)
+        let cell: PersonPhotoCell = collectionView.dequeueCell(for: indexPath)
+        cell.personImage = #imageLiteral(resourceName: "man-in-red-jacket-1681010")
```
Kilo-Loco, line 44: I don’t think this has been addressed in the project yet but we shouldn’t use image literals as they can cause the app to crash if the asset is removed. Can we use UIImage(named:) instead?

Say Their Names/Source/Controller/Home/Person/PersonDetailsController.swift
```diff
@@ -41,12 +40,13 @@ extension PersonDetailsController: UICollectionViewDataSource, UICollectionViewD
...
         return cell
     }
     
     func collectionView(_ collectionView: UICollectionView, layout collectionViewLayout: UICollectionViewLayout, minimumLineSpacingForSectionAt section: Int) -> CGFloat {
-        return 18
+        return 15
```
Kilo-Loco, line 49: Wondering if we should come up with a convention to handle magic numbers

Say Their Names/Source/Utils/Protocols/Reusable.swift
```diff
@@ -0,0 +1,37 @@
...
+//  Say Their Names
+//
+//  Created by Leonardo Garcia  on 02/06/20.
+//  Copyright © 2020 Franck-Stephane Ndame Mpouli. All rights reserved.
+//
+
+import UIKit
+
+protocol Reusable {
+    static var reuseIdentifier: String { get }
```
Kilo-Loco, line 12: Nice! I do this myself! 🤩

Say Their Names/Source/Utils/Protocols/Reusable.swift
```diff
@@ -0,0 +1,37 @@
...
+extension Reusable {
+    static var reuseIdentifier: String {
+        return String(describing: self)
+    }
+}
+
+extension UITableViewCell: Reusable { }
+extension UICollectionViewCell: Reusable { }
+
+extension UICollectionView {
```
Kilo-Loco, line 24: Can we move this to a separate extensions file? `UICollectionView++`

Say Their Names/Source/View/Home/Person/PersonPhotoCell.swift
```diff
@@ -8,19 +8,63 @@
...
 
-class PersonPhotoCell: UICollectionViewCell {
-
-    override func awakeFromNib() {
-        super.awakeFromNib()
-        // Initialization code
+final class PersonPhotoCell: UICollectionViewCell {
+    
+    var personImage: UIImage? {
+        didSet {
```
Kilo-Loco, line 14: Latest s updates to Swift 5 suggest that we shouldn’t be using `didSet` unless we are specifically working with logic that involves `oldValue`.

Also, we will be using the `ImageService` to handle threading and setting the image

Say Their Names/Source/View/Home/Person/PersonPhotoCell.swift
```diff
@@ -8,19 +8,63 @@
...
+    
+    @available(*, unavailable)
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
     }
+    
+    private func setup() {
+        addSubview(personImageView)
+        
+        personImageView.topAnchor.constraint(equalTo: topAnchor).isActive = true
```
Kilo-Loco, line 46: This one ran away from the rest

Say Their Names/Source/View/Home/Person/PersonPhotoCell.swift
```diff
@@ -8,19 +8,63 @@
...
+// MARK: Design Constants
+extension PersonPhotoCell {
+    private static var aspectRatio: CGFloat { 0.7 }
 }
 
 extension PersonPhotoCell {
     static func size(_ collectionView: UICollectionView) -> CGSize {
         let size = collectionView.bounds.size
-        let width = (size.height-24)-1
+        let width = size.height * aspectRatio
```
Kilo-Loco, line 67: Much cleaner 😁

Say Their Names/Source/View/Home/Person/PersonPhotoCell.swift
leonardo-garcia: Nice catch! I was doing some testing and forgot to get it back into the array

Say Their Names/Source/View/Home/Person/PersonPhotoCell.swift
leonardo-garcia: I wasn't aware of this, thanks. Then, should personImageView be exposed to be internal?

leonardo-garcia: I got it, you meant I shouldn't be sending it to the main queue, since it is responsibility of `ImageService`. I've update it. Although I find safer to send it to the main queue in the actual implementation of the change of the image, I think it is less error-prone than expecting the call to the function to be made in the correct queue

Say Their Names/Source/View/Home/Person/PersonPhotoCell.swift
```diff
@@ -8,19 +8,57 @@
...
+        addSubview(personImageView)
+                
+        let constraints = [
+            personImageView.topAnchor.constraint(equalTo: topAnchor),
+            personImageView.leadingAnchor.constraint(equalTo: leadingAnchor),
+            personImageView.trailingAnchor.constraint(equalTo: trailingAnchor),
+            personImageView.bottomAnchor.constraint(equalTo: bottomAnchor)
+        ]
+        
+        NSLayoutConstraint.activate(constraints)
```
Mackis, line 45: No need to declare single-use constants here 

```
NSLayoutConstraint.activate([
            personImageView.topAnchor.constraint(equalTo: topAnchor),
            personImageView.leadingAnchor.constraint(equalTo: leadingAnchor),
            personImageView.trailingAnchor.constraint(equalTo: trailingAnchor),
            personImageView.bottomAnchor.constraint(equalTo: bottomAnchor)
        ])
```

Say Their Names/Source/Utils/Extentions/UICollectionView++.swift
```diff
@@ -0,0 +1,26 @@
...
+    /// Dequeues a generic cell from the collection view using the reuseIdentifier coming from the `Reusable` protocol.
+    /// Note that the cell must be registered before into the collection view with the same reuseIdentifier.
+    func dequeueCell<T: UICollectionViewCell>(for indexPath: IndexPath) -> T {
+        guard let cell = dequeueReusableCell(withReuseIdentifier: T.reuseIdentifier, for: indexPath) as? T else {
+            fatalError("You need to register cell of type `\(T.reuseIdentifier)`")
+        }
+        
+        return cell
+    }
+}
```
unnamedd, line 26: 👏👏👏👏

Say Their Names/Source/Utils/Protocols/Reusable.swift
```diff
@@ -0,0 +1,20 @@
...
+/// Provides a default `reuseIdentifier` to the entities that conform to it.
+protocol Reusable {
+    static var reuseIdentifier: String { get }
+}
+
+extension Reusable {
+    static var reuseIdentifier: String {
+        return String(describing: self)
+    }
+}
```
unnamedd, line 20: This is the best way so far to have an identifier!

Say Their Names/Source/View/Home/Person/PersonPhotoCell.swift
hybridcattt: it helps with type checking time, especially if we add type annotation to the `constraints` variable. not a big deal with 4 constraints, but if it's over 10, usually makes a difference

hybridcattt:
hmm I see that was probably intentional to have it named `UICollectionView++.swift`. Never seen that kind of naming, but I get it. So not changing then 👍

unnamedd:
> hmm I see that was probably intentional to have it named `UICollectionView++.swift`. Never seen that kind of naming, but I get it. So not changing then 👍

Plus signs is used to be used for extensions (or categories when we talk about Objective-C), in this case, I'd use `++` only if I'm being lazy, in this case, we should avoid this kind of filename pattern to keep consistency with what we have inside of the file. For that specific case, we should name it: `UICollectionView+ReusableCell.swift` or something on this way.

Say Their Names/Source/Utils/Extentions/UICollectionView++.swift
FranckNdame, line 26: Nice!

Say Their Names/Source/View/Home/Person/PersonPhotoCell.swift
```diff
@@ -8,19 +8,57 @@
...
+    private func setup() {
+        addSubview(personImageView)
+                
+        let constraints = [
+            personImageView.topAnchor.constraint(equalTo: topAnchor),
+            personImageView.leadingAnchor.constraint(equalTo: leadingAnchor),
+            personImageView.trailingAnchor.constraint(equalTo: trailingAnchor),
+            personImageView.bottomAnchor.constraint(equalTo: bottomAnchor)
+        ]
+
```
FranckNdame, line 44: Added a handy extension. (UIView++)
I'll change line 39-42 to `personImageView.fillSuperview()`

PR #29, opened 2020-06-03
Home controller programmatic layout

jaywardell:
factored a ServiceClient protocol out of BaseViewController
created a new ServiceClientController that conforms to ServiceClient but doesn't init from a nib
HomeController now is a subclass of ServiceClientViewController
Created HomeView, a UIView subclass, that contains and vends collections views and nav bar
HomeViewController loads a HomeView in loadView and uses HomeView's subviews, no longer loads from a xib
all views are slayed out programmatically with Autolayout
old xib is removed


Say Their Names/Source/Controller/Home/HomeView.swift
```diff
@@ -0,0 +1,160 @@
...
+//
+//  Created by Joseph A. Wardell on 6/2/20.
+//  Copyright © 2020 Franck-Stephane Ndame Mpouli. All rights reserved.
+//
+
+import UIKit
+
+class HomeView: UIView {
+
+    var customNavBar: UIView!
```
Mackis, line 13: My suggestion would be to keep view configuration close to where it's declared. 
Then we can have init code separately to view setup code. This keeps initializers smaller and makes it easier to debug specific views. 

Also gets rid of the bang(!) 

```
lazy var customNavBar: UIView = {
let customNavBar = UIView()
customNavBar.backgroundColor = .black
return customNavBar
}
```

Say Their Names/Source/Controller/Home/HomeView.swift
```diff
@@ -0,0 +1,160 @@
...
+            peopleCollectionView.bottomAnchor.constraint(equalTo: collections.bottomAnchor)
+        ])
+    }
+    
+    // MARK:- Constants
+    static let CustomNavBarHeight : CGFloat = 70
+    static let PeopleCollectionViewHeight : CGFloat = 70
+    static let SearchButtonSize : CGFloat = 45
+    static let CustomNavBarMargin : CGFloat = 16
+    static let SeparatorHeight : CGFloat = 1
```
Mackis, line 156: @Kilo-Loco mentioned in another diff that we will probably need to extract such values into a separate file. Maybe we can create a `Theme.swift` or `Styles.swift`.

Say Their Names/Source/Controller/Home/HomeView.swift
```diff
@@ -0,0 +1,160 @@
...
+
+import UIKit
+
+class HomeView: UIView {
+
+    var customNavBar: UIView!
+    var locationCollectionView: UICollectionView!
+    var peopleCollectionView: UICollectionView!
+    var searchButton: UIButton!
+    var separator : UIView!
```
unnamedd, line 17:
```suggestion
    var separatorView : UIView!
```

Say Their Names/Source/Controller/Home/HomeView.swift
```diff
@@ -0,0 +1,160 @@
...
+//  HomeView.swift
+//  Say Their Names
+//
+//  Created by Joseph A. Wardell on 6/2/20.
+//  Copyright © 2020 Franck-Stephane Ndame Mpouli. All rights reserved.
+//
+
+import UIKit
+
+class HomeView: UIView {
```
unnamedd, line 11: If you are not going to subclass it, then it is final.

```suggestion
final class HomeView: UIView {
```

Say Their Names/Source/Controller/Base/ServiceClientViewController.swift
```diff
@@ -0,0 +1,23 @@
...
+//  ServiceClientViewController.swift
+//  Say Their Names
+//
+//  Created by Joseph A. Wardell on 6/2/20.
+//  Copyright © 2020 Franck-Stephane Ndame Mpouli. All rights reserved.
+//
+
+import UIKit
+
+class ServiceClientViewController : UIViewController, ServiceClient  {
```
unnamedd, line 11: It is not really clear for me, the intention of this ServiceClientViewController. Your plan is to make every ViewController by default have access to the `ServiceClient`? Would you explain it?

Say Their Names/Source/Controller/Home/HomeView.swift
unnamedd: I guess we should use the full name instead of abbreviated names. Instead of: `Nav`, use `Navigation`.

Say Their Names/Source/Controller/Home/HomeView.swift
unnamedd: I completely agree to have all the ViewCode separated into a different file.

Say Their Names/Source/Controller/Base/ServiceClient.swift
```diff
@@ -0,0 +1,16 @@
...
+//  ServiceClient.swift
+//  Say Their Names
+//
+//  Created by Joseph A. Wardell on 6/2/20.
+//  Copyright © 2020 Franck-Stephane Ndame Mpouli. All rights reserved.
+//
+
+import UIKit
+
+protocol ServiceClient {
```
hybridcattt, line 11: There is `ServiceReferring` protocol already which is exactly the same 😅  you could use that (renaming it if you like `ServiceClient` is a better name)

Say Their Names/Source/Controller/Base/ServiceClientViewController.swift
hybridcattt, line 11: We already have `BaseViewController`. I'm personally not in favour of base classes unless absolutely necessary, but even if we do have them, let's have one, not two :)

Say Their Names/Source/Controller/Home/HomeController.swift
```diff
@@ -41,6 +49,11 @@ class HomeController: BaseViewController {
         // Select first location by default
         let selectedIndexPath = IndexPath(item: 0, section: 0)
         locationCollectionView.selectItem(at: selectedIndexPath, animated: false, scrollPosition: .centeredVertically)
+        
+        // hide the major parts of the homeview
+        // so the launch screen will appear
+        homeView.hideable.forEach { $0.isHidden = true }
```
hybridcattt, line 55: this shouldn't be needed anymore after latest fixes to launch screen code 🤔

Say Their Names/Source/Controller/Home/HomeController.swift
```diff
@@ -135,6 +152,7 @@ extension HomeController {
 
         let completion = {
             self.revealTabBar()
+            self.homeView.hideable.forEach { $0.isHidden = false }
```
hybridcattt, line 155: this shouldn't be needed anymore after latest fixes to launch screen code - here too

Say Their Names/Source/Controller/Home/HomeView.swift
hybridcattt: to build up on @Mackis 's suggestion, we can even do 
```
let customNavBar: UIView = {
   let customNavBar = UIView()
   customNavBar.backgroundColor = .black
   return customNavBar
}()
```
to avoid unnecessary mutability.
The improvement can be done as a follow-up I think, just make sure to create a Trello card for it 🙏

Say Their Names/Source/Controller/Home/HomeView.swift
```diff
@@ -0,0 +1,160 @@
...
+    override func didMoveToSuperview() {
+        super.didMoveToSuperview()
+        
+        createLayout()
+        
+    }
+    
+    private var hasLayedOutSubviews = false
+    private func createLayout() {
+        guard !hasLayedOutSubviews else { return }
```
hybridcattt, line 71: `updateConstraints()` override is a more appropriate place for setting up constraints. the end result would be pretty much the same :)

Say Their Names/Source/Controller/Base/ServiceClientViewController.swift
unnamedd: I completely agree with @hybridcattt, we must avoid as much we can to have "bases".

Say Their Names/Source/Controller/Home/HomeView.swift
jaywardell: It's easy enough to do in this PR.  Thanks for the nudge.

Say Their Names/Source/Controller/Home/HomeView.swift
jaywardell: Agreed that this is a good idea. I've added a Trello card for this.

Say Their Names/Source/Controller/Home/HomeView.swift
jaywardell: done

Say Their Names/Source/Controller/Home/HomeView.swift
jaywardell: Typically, if I see the suffix "View" at the end of a property name, it suggests to me that the property lets me view that thing (e.g. tableView, personView, etc.).  This is just a separator meant to visually denote where one collection view ends and the next begins.  I think separator is the right name.

Say Their Names/Source/Controller/Home/HomeView.swift
jaywardell: done

Say Their Names/Source/Controller/Base/ServiceClient.swift
jaywardell: Thanks.  I didn't even see that.  I've switched to using ServiceReferring. This allowed me to get rid of the ServiceClientViewController.

Say Their Names/Source/Controller/Base/ServiceClientViewController.swift
jaywardell: Okay, done.  I now just conform HomeViewController to ServiceReferring.  

I was just trying to follow the conventions of the project as I saw it.  My plan was simply to gradually move all view controllers over to ServiceClientViewController subclasses as we switched from xib to programmatic UI (since BaseViewController as written won't let us init without using a nib). 

If the consensus is that we don't need a base view controller, I'm very happy with that.

Say Their Names/Source/Controller/Home/HomeView.swift
jaywardell: Agreed.  Done.  It also let me find and fix a weird layout bug when moving from portrait to landscape. Thanks.

Say Their Names/Source/Controller/Base/ServiceClientViewController.swift
jaywardell: see my conversation with hybridcattt below.  I've removed it.

Say Their Names/Source/Controller/Home/HomeController.swift
jaywardell: I'm willing to believe you. With the branch I have, it is needed. Can we merge this into development and then see?

Say Their Names/Source/Controller/Home/HomeController.swift
jaywardell: see note for above.

Say Their Names/Source/Controller/Base/ServiceClientViewController.swift
hybridcattt: That sounds great! I don’t look too closely at the BaseViewController. If you see any inconsistencies there with regards to programmatic UI, feel free to fix it or create a Trello card 😊

Say Their Names/Source/Controller/Home/HomeController.swift
hybridcattt: Fixed this during merge 🙂

hybridcattt: (This part of the code was moved to the tabbar subclass)

Say Their Names/Source/Controller/Home/HomeView.swift
hybridcattt: 🙌

PR #30, opened 2020-06-03
UI updates to PersonCell

thewhitewood:
- Updated layout
- Added outlets
- Refactored PersonCell
- Fixed incorrect case for bookmark asset


Say Their Names/Source/View/Home/Person/PersonCell.swift
```diff
@@ -8,11 +8,28 @@
...
+    // MARK:- Lifecycle
+    
     override func awakeFromNib() {
         super.awakeFromNib()
-        // Initialization code
     }
 
+    // MARK:- Handlers
+        
+    @IBAction func bookmarkButtonDidTouchUpInsde(_ sender: Any) {
```
unnamedd, line 32: You don't need to specify everything you are doing in the name of the method. 
First, this is a function, we know that functions are actions.
Second, the sender itself has the type of element used here (which must be UIButton instead of Any).
Third, if you don't want to make it available outside of the class, it must be private. Most of the time, this kind of method will be private.

```suggestion
    @IBAction private func bookmark(_ sender: UIButton) {
```

Say Their Names/Source/View/Home/Person/PersonCell.swift
```diff
@@ -8,11 +8,28 @@
 
 import UIKit
 
-class PersonCell: UICollectionViewCell {
+final class PersonCell: UICollectionViewCell {
 
+    @IBOutlet weak var nameLabel: UILabel!
+    @IBOutlet weak var imgView: UIImageView!
+    @IBOutlet weak var dobLabel: UILabel!
+    @IBOutlet weak var bookmarkButton: UIButton!
```
unnamedd, line 16: Always make them private.
Avoid raw variables, e.g: `imgView`, `dobLabel`, always use the fullname. `imageView` or `birthdayLabel`. In this case, I guess the `imageView`, will be the user's profile picture, then, it might be `profileImageView`.

```suggestion
    @IBOutlet weak private var nameLabel: UILabel!
    @IBOutlet weak private var profileImageView: UIImageView!
    @IBOutlet weak private var birthdayLabel: UILabel!
    @IBOutlet weak private var bookmarkButton: UIButton!
```

Say Their Names/Source/View/Home/Person/PersonCell.swift
```diff
@@ -8,11 +8,28 @@
...
+    
+    var data: Person? {
+        didSet {
+            //guard let data  = data else {return}
+        }
+    }
```
unnamedd, line 22: In the future, you might need to set up the view as soon as you receive the object in this variable (which is better to rename it to `person` instead of `data`), but until there, you must avoid having unused commented code.

```suggestion
    var person: Person?
```

thewhitewood: Yeh a lot of this was taken from previous implementations in the project - will wait for direction to be made on coded layouts before updating.

PR #28, opened 2020-06-03
Ios create custom cv data source

iJACD:
Implemented by @cdowney102  and I.

Created custom collection view data source classes for person and location cell in order to reduce view controller clutter and abstract out some responsibilities. 
Updated HomeController to use custom data source. 
Added dummy data to populate and configure cells.
Added appendPeople function to support paging logic when implemented.
Added configure(with: ) function to location and person cell.

Future work:
Will need to implement Kilos image downloader for cell image source.

PR #37, opened 2020-06-03
Post-merge cleanup

hybridcattt:
Accidentally left couple things commented out during merge of #28. Fixing that now + the tag approach was confusing to me, so improved that too.

### DIFF
--- a/Say Their Names/Source/Controller/Home/CollectionViewDataSource/PersonCollectionViewDataSource.swift
+++ b/Say Their Names/Source/Controller/Home/CollectionViewDataSource/PersonCollectionViewDataSource.swift
@@ -37,7 +37,6 @@ extension PersonCollectionViewDataSource: UICollectionViewDataSource {
     }
     
     func collectionView(_ collectionView: UICollectionView, viewForSupplementaryElementOfKind kind: String, at indexPath: IndexPath) -> UICollectionReusableView {
-        if collectionView.tag == 0 { return UICollectionReusableView() }
         let headerView = collectionView.dequeueReusableSupplementaryView(ofKind: UICollectionView.elementKindSectionHeader, withReuseIdentifier: PersonHeaderCell.headerIdentifier, for: indexPath) as! PersonHeaderCell
         return headerView
     }
--- a/Say Their Names/Source/Controller/Home/HomeController.swift
+++ b/Say Their Names/Source/Controller/Home/HomeController.swift
@@ -57,6 +57,7 @@ class HomeController: UIViewController, ServiceReferring {
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
         // Select first location by default
+        // FIXME: can have multiple selected. need one source-of-truth here.
         let selectedIndexPath = IndexPath(item: 0, section: 0)
         locationCollectionView.selectItem(at: selectedIndexPath, animated: false, scrollPosition: .centeredVertically)
     }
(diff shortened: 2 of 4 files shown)


Say Their Names/Source/Controller/Home/CollectionViewDataSource/PersonCollectionViewDataSource.swift
```diff
@@ -37,7 +37,6 @@ extension PersonCollectionViewDataSource: UICollectionViewDataSource {
     }
     
     func collectionView(_ collectionView: UICollectionView, viewForSupplementaryElementOfKind kind: String, at indexPath: IndexPath) -> UICollectionReusableView {
-        if collectionView.tag == 0 { return UICollectionReusableView() }
         let headerView = collectionView.dequeueReusableSupplementaryView(ofKind: UICollectionView.elementKindSectionHeader, withReuseIdentifier: PersonHeaderCell.headerIdentifier, for: indexPath) as! PersonHeaderCell
```
evilpenguin, line 40: Why are we casting here? This isn't needed unless we are doing things with `PersonHeaderCell`

Say Their Names/Source/Controller/Home/HomeController.swift
```diff
@@ -119,42 +120,37 @@ class HomeController: UIViewController, ServiceReferring {
...
-        let width = collectionView.frame.width - 32
-        return collectionView.tag == 0 ? .zero : .init(width: width, height: 170)
+        if collectionView === locationCollectionView {
+            return CGSize.zero
+        }
+        else if collectionView === peopleCollectionView {
+            let width = collectionView.frame.width - 32
+            return CGSize(width: width, height: 170)
+        }
+        else {
```
evilpenguin, line 130: I know the `else {}` wrapping the final return is readability but not required.

Say Their Names/Source/Controller/Home/HomeController.swift
```diff
@@ -119,42 +120,37 @@ class HomeController: UIViewController, ServiceReferring {
...
-        return collectionView.tag == 0 ? locationCellSize : peopleCellSize
+        
+        if collectionView === locationCollectionView {
+            return CGSize(width: 103, height: 36)
+        }
+        else if collectionView === peopleCollectionView {
+            let width = collectionView.frame.width / 2 - 24
+            return CGSize(width: width, height: 300)
+        }
+        else {
```
evilpenguin, line 144: Same here.

Say Their Names/Source/Controller/Home/HomeController.swift
```diff
@@ -119,42 +120,37 @@ class HomeController: UIViewController, ServiceReferring {
...
+            return CGSize.zero
+        }
     }
     
     func collectionView(_ collectionView: UICollectionView, didSelectItemAt indexPath: IndexPath) {
-        if collectionView.tag == 0 {
-            // Location CollectionView
-            
-        } else {
+        if collectionView === locationCollectionView {
```
evilpenguin, line 150: Do we really need code that does nothing? Can we remove this?

evilpenguin: Maybe a simple comment for now.

evilpenguin:
Saw comment on email about comment vs imple.

hybridcattt:
Here I changed it so code follows different path depending on which collection view instance it is. And if we add a third collection view, so it wouldn't break. 
If there wasn't a third possible option, then yeah it's just a matter of taste I think :)

PR #38, opened 2020-06-03
Refactor Donations to be programmatic UI

kingsqd:
In this PR I am refactoring Donations controller to use programmatic UI

Screen shot: 
Simulator Screen Shot - iPhone 11 - 2020-06-03 at 12 35 54


Say Their Names/Source/Controller/Donations/DonationsViewCode.swift
```diff
@@ -0,0 +1,53 @@
...
+//  Created by Hakeem King on 6/3/20.
+//  Copyright © 2020 Franck-Stephane Ndame Mpouli. All rights reserved.
+//
+
+import UIKit
+
+/// The UI for Donations
+final class DonationsViewCode: UIView {
+    
+    private lazy var donationsLabel: UILabel = {
```
hybridcattt, line 14: tiny thing - this can be a let, it's anyway first accessed from `init` so not really getting any laziness 

```suggestion
    private let donationsLabel: UILabel = {
```

Say Their Names/Source/Controller/Donations/DonationsViewCode.swift
```diff
@@ -0,0 +1,53 @@
...
+        let label = UILabel()
+        label.text = "Donations"
+        label.textAlignment = .center
+        
+        return label
+    }()
+    
+    init() {
+        super.init(frame: .zero)
```
hybridcattt, line 23:
```suggestion
    init(frame: CGRect) {
        super.init(frame: frame)
```

if you do it like that, still can call `init()` but also `init(frame: x)`

Say Their Names/Source/Controller/Donations/DonationsViewCode.swift
kingsqd: I love these small comments. I am an Obj-c dev, so the small things in Swift I miss.

Say Their Names/Source/Controller/Donations/DonationsController.swift
```diff
@@ -9,8 +9,18 @@
...
+    required init(service: Service, shouldInitWithNib: Bool) {
+        super.init(service: service, shouldInitWithNib: false)
+    }
+    
+    required init?(coder: NSCoder) {
+        fatalError()
+    }
+    
+    override func loadView() {
+        view = viewCode
```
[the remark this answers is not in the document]
hybridcattt: `viewCode` is already used elsewhere so I think it's ok to keep. Let's finish the transition to programmatic, and then we could unify all names to the one we like most

PR #43, opened 2020-06-04
Ios date formatter service

kingsqd:
This PR is creating a date formatting service per this ticket : https://trello.com/c/CjE2733E

Currently this formatter supports a DOB format and a hour minute format. I am open to changing the formats or adding more, but I just wanted to begin with these two to get a service up and running. 

We use a singleton instance to avoid creating new instances across the app.

### DIFF
--- a/Say Their Names/Source/Service/DateFormatter/DateFormatterService.swift
+++ b/Say Their Names/Source/Service/DateFormatter/DateFormatterService.swift
@@ -15,8 +15,6 @@ protocol DateFormatterType {
 extension DateFormatter: DateFormatterType { }
 
 class DateFormatterService {
-
-    static let shared = DateFormatterService()
     
     /// To keep thread safe, designate this queue for searching cached formatters.
     let dateFormattersQueue = DispatchQueue(label: "com.stn.date.formatter.queue")
@@ -42,10 +40,10 @@ class DateFormatterService {
 
     // MARK: - Date Of Birth
 
-    func formatDOBDate(_ date: Date) -> String {
-        let dateFormatter = cachedDateFormatter(withFormat: "y/MM/dd @ HH:mm")
+    func formatYearMonthDayDate(_ date: Date) -> String {
+        let dateFormatter = cachedDateFormatter(withFormat: "y/MM/dd")
         let formattedDate = dateFormatter.string(from: date)
-        return ("Date of birth: \(formattedDate)")
+        return ("Date: \(formattedDate)")
     }
     
     // MARK: - Hour minute
--- a/Say Their Names/Source/Service/DateFormatter/DateFormatterService.swift
+++ b/Say Their Names/Source/Service/DateFormatter/DateFormatterService.swift
@@ -38,7 +38,7 @@ class DateFormatterService {
         }
     }
 
-    // MARK: - Date Of Birth
+    // MARK: - Year month day
 
     func formatYearMonthDayDate(_ date: Date) -> String {
         let dateFormatter = cachedDateFormatter(withFormat: "y/MM/dd")
--- a/Say Their Names/Source/Service/DateFormatter/DateFormatterService.swift
+++ b/Say Their Names/Source/Service/DateFormatter/DateFormatterService.swift
@@ -1,0 +1,56 @@
+//
+//  STNDateFormatter.swift
+//  Say Their Names
+//
+//  Created by Hakeem King on 6/3/20.
+//  Copyright © 2020 Franck-Stephane Ndame Mpouli. All rights reserved.
+//
+
+import Foundation
+
+protocol DateFormatterType {
+    func string(from date: Date) -> String
+}
+
+extension DateFormatter: DateFormatterType { }
+
+class DateFormatterService {
+
+    // MARK: - Shared
+
+    static let shared = DateFormatterService()
+    
+    // MARK: - Queue
+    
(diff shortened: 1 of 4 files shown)


rodericj:
My recommendation would be to use dependency injection instead of a singleton. I’m happy to talk about it for days if anyone wants to know more.

Say Their Names/Source/Service/DateFormatter/DateFormatterService.swift
```diff
@@ -0,0 +1,58 @@
...
+
+protocol DateFormatterType {
+    func string(from date: Date) -> String
+}
+
+extension DateFormatter: DateFormatterType { }
+
+class DateFormatterService {
+
+    static let shared = DateFormatterService()
```
evilpenguin, line 19: We are going with dependency injection for the application. Can you please add this to `Service`

Say Their Names/Source/Service/DateFormatter/DateFormatterService.swift
```diff
@@ -0,0 +1,58 @@
...
+            
+            dateFormatters[key] = dateFormatter
+            
+            return dateFormatters[key]!
+        }
+    }
+
+    // MARK: - Date Of Birth
+
+    func formatDOBDate(_ date: Date) -> String {
```
evilpenguin, line 45: We also have `date_of_incident`. Can we make this more generic?

Say Their Names/Source/Service/DateFormatter/DateFormatterService.swift
```diff
@@ -0,0 +1,58 @@
...
+        let dateFormatter = cachedDateFormatter(withFormat: "y/MM/dd @ HH:mm")
```
evilpenguin, line 46: I don't think our dates have time? https://saytheirnames.dev/api/people

kingsqd:
@rodericj I am open to this. would love to hear how we could benefit from using dependency injection over singleton.

Say Their Names/Source/Service/Service.swift
```diff
@@ -12,6 +12,7 @@ import UIKit
 final class Service {
     lazy private(set) var navigator = Navigator(service: self)
     lazy private(set) var image = ImageService()
+    lazy private(set) var dateFormatter = DateFormatterService.shared
```
evilpenguin, line 15: Lets keep this a single instance. We're trying to avoid any possible singletons.

rodericj:
Looks like @evilpenguin had a few of the same thoughts and unfortunately this may mean you’ll need to rewrite a bit of this but I don’t think it’ll be hard for you. 

Dependency injection is fantastic for testing (yes I see that you have tests here and I love it). If you think about how we would test some of these things without dependency injection that depend on date formatters you end up with a constructor that is mostly empty, you create the object, call “showMeSomeStringThatIncludesADate() ->”

That’s pretty easy to verify with a unit test, but the world has now changed. You now have some cache thing that’s changed with the newly created instance of the dateFormatter. So is that something we need to test also? Is it really relevant to our UiViewWitgADateRenderedAsALavelORSomething? It’s kind of not. 

With dep inj we now have to deal with a longer constrictor which takes in our date formatter. But when we test our func, there are no side effects. If we call it again, the world is the same before and after each call!

Ok so admittedly a date formatter isn’t the most earth shattering object and the world isn’t THAT different now that we have some static dictionary with one more element. But for consistency sake you gotta avoid the bad pattern. Because of you don’t draw the line at this date formatter, will you draw the line at the data store handler? Or the network service?

Anyway this was just a quick example. Let me know if you need clarification or more resources or if you have follow ups.

evilpenguin:
Please ignore the revert.

kingsqd:
@rodericj Yes, this completely makes sense now. Thanks for this detailed explanation, on this. But I would like to take you up on your offer of providing more resources on the topic to better understand for my own learnings. 

Thanks !

rodericj:
@kingsqd I've not forgotten about your request @kingsqd. I'm trying to find some good links and I'm starting with wikipedia. https://en.wikipedia.org/wiki/Dependency_injection. They have a good analogy of driving. You drive a car, and it really doesn't matter if it's a gas car, electric car, Flintstone car. You use the same interface (gas pedal, brake pedal, steering wheel) to control this thing, and what happens under the hood is not really relevant to you. Someone could swap out your gas engine for a battery electric engine and it should still work the same, gas makes you go faster, brake makes you slow down, wheel turns you. If you were to swap out a "test audio engine" which just made a loud noise when you stepped on the gas that'd be something that you kind of just have to deal with.

So our dependency injection uses something that conforms to `Servicing`. Right now there is just `Service` so when you call `service.network.getPeople()` you as a smart person would expect a network request to go out to some web service and return some people. In an automated testing system on CI somewhere with slow internet, you can't depend on the network actually returning people, so you mock this with a network client that maybe just returns `["James", "Robert"] so you can test a collection view of 2 people. Dependency injection makes this super super easy.

rodericj:
There are so many videos out there and they all look familiar, but honestly, just play around with this and I'll start pointing you towards tests as I write them. You might see how awesome it is then.

https://www.youtube.com/watch?v=IKD2-MAkXyQ

rodericj:
He does a much better job of explaining things https://www.swiftbysundell.com/articles/avoiding-singletons-in-swift @kingsqd

PR #50, opened 2020-06-04
Add image loading library 

juanfolivera:
**Please have in mind that images were added just for demonstration**

Added Kingfisher library for image loading and caching

ezgif com-video-to-gif


Say Their Names/Source/View/Home/Person/PersonPhotoCell.swift
```diff
@@ -35,10 +36,15 @@ final class PersonPhotoCell: UICollectionViewCell {
...
     }
     
-    func setImage(_ image: UIImage?) {
-        personImageView.image = image
+    func setImage(withUrlString url: String) {
+      personImageView.kf.indicatorType = .activity
+      personImageView.kf.setImage(
+        with: URL(string: url),
+        placeholder: UIImage(named: "image-placeholder")
+      )
```
hybridcattt, line 47: I see this could be made into a reusable function that takes image url and placeholder image as parameters - to reduce duplicated code

juanfolivera: Agree, I just thought I was overthinking it at the time, that's why I left it as it is, but I'll make those updates

Say Their Names/Source/View/Home/Person/PersonCell.swift
```diff
@@ -79,10 +80,11 @@ final class PersonCell: UICollectionViewCell {
     func configure(with person: Person) {
         let df = DateFormatter()
         df.dateFormat = "dd.MM.yyyy"
-        
-        profileImageView.image = UIImage(named: "man-in-red-jacket-1681010")
-        // ^Update to default place holder image when resource is added.
-        // TO-DO: Use Kilo's ImageDownloader here to DL image
+        profileImageView.kf.indicatorType = .activity
+        profileImageView.kf.setImage(
+          with: URL(string: person.media[0]), //TODO: Add thumbnail to person model?
```
Kilo-Loco, line 85: This seems dangerous because we can end up with an index out of range error if there aren't any media for this person. There are a couple ways to fix this; here's one:
```suggestion
          with: URL(string: person.media.first ?? ""), //TODO: Add thumbnail to person model?
```

rodericj: if I'm not mistaken this can be a linter warning. Might be something to look into.

Kilo-Loco:
The majority of this PR was really good. Why close?

juanfolivera:
@Kilo-Loco Sorry too many conflicts due to the name changing, i'm recreating a new one with all the changes suggested here

PR #53, opened 2020-06-04
Fix linter warnings for forced unwrapping

rodericj:


rodericj:
Thanks for the fast 👍

SayTheirNames/Source/Components/MainTabBarController.swift
```diff
@@ -64,7 +64,10 @@ class MainTabBarController: UITabBarController, UITabBarControllerDelegate {
         tabBar.barTintColor = defaultBarTint
         tabBar.tintColor = defaultTint
         tabBar.unselectedItemTintColor = defaultUnselectedTint
-        UITabBarItem.appearance().setTitleTextAttributes([NSAttributedString.Key.font: UIFont(name: "Karla-Regular", size: 11)!], for: .normal)
+        guard let karlaFont = UIFont(name: "Karla-Regular", size: 11) else {
```
Kilo-Loco, line 67: We should try to use our semantic fonts when possible to keep consistency
```suggestion
        let karlaFont = UIFont.STN.tabButtonTitle
```

rodericj: Done: https://github.com/Say-Their-Name/say-their-names-ios/pull/55

rodericj: Done in #55

PR #58, opened 2020-06-04
Added Alamofire and Networking

evilpenguin:


SayTheirNames/Source/Service/Networking/NetworkRequestor.swift
```diff
@@ -0,0 +1,44 @@
...
+//  Created by evilpenguin on 6/3/20.
+//  Copyright © 2020 Franck-Stephane Ndame Mpouli. All rights reserved.
+//
+
+import Foundation
+import Alamofire
+
+// MARK: - NetworkRequestor
+
+class NetworkRequestor: NSObject, ServiceReferring {
```
hybridcattt, line 14:
```suggestion
final class NetworkRequestor {
```
I think it wouldn't need a service, if it just makes requests and parses responses 🤔 the reversed, `Service` holds the requestor. Otherwise there's circular dependencies...

SayTheirNames/Source/Controller/Home/HomeController.swift
```diff
@@ -74,28 +73,16 @@ class HomeController: UIViewController, ServiceReferring {
...
         locationsDataSource.setLocations(locations)
-        peopleDataSource.setPeople(people)
-        
+
+        // FIXME: This should be setup in a better place, for now this loads out data
+        self.service.network.fetchPeople { [weak self] (p) in
+            guard let people = p?.all else { return }
+            
+            self?.peopleDataSource.setPeople(people)
+            self?.peopleCollectionView.reloadData()
```
hybridcattt, line 83: 🎉 🎉 🎉

SayTheirNames/Source/Model/Donation.swift
```diff
@@ -1,19 +1,40 @@
 //
 //  Donation.swift
-//  SayTheirNames
+//  Say Their Names
```
hybridcattt, line 3: revert back? (also few other places)

SayTheirNames/Source/Model/Image.swift
```diff
@@ -0,0 +1,19 @@
...
+//
+//  Created by evilpenguin on 6/3/20.
+//  Copyright © 2020 Franck-Stephane Ndame Mpouli. All rights reserved.
+//
+
+import Foundation
+
+public struct Image: Decodable {
+    var id: Int
+    var personId: Int
```
hybridcattt, line 13: by swift capitalisation rules it should be `personID`, `requestURL`, etc

SayTheirNames/Source/Model/Person.swift
```diff
@@ -1,28 +1,63 @@
...
+    var country: String
     var bio: String
     var context: String
-    var donations: [Donation]
-    var petitions: [Petition]
+    var images: [Image]
+    var donations: Donations?
+    var petitions: Petitions?
+    var media: [Media]?
+    var socialMedia: [SocialMedia]?
```
hybridcattt, line 26: do we need an optional? could array just be empty, or is it because of the way `Codable` works?

SayTheirNames/Source/Model/Person.swift
```diff
@@ -1,28 +1,63 @@
...
+        self.images = try values.decodeIfPresent([Image].self, forKey: .images) ?? []
+        
+        self.donations = try values.decodeIfPresent(Donations.self, forKey: .donations)
+        self.petitions = try values.decodeIfPresent(Petitions.self, forKey: .petitions)
+        self.media = try values.decodeIfPresent([Media].self, forKey: .media)
+        self.socialMedia = try values.decodeIfPresent([SocialMedia].self, forKey: .socialMedia)
+    }    
+}
+
+public struct Persons: Decodable {
```
hybridcattt, line 56: Could this be called smth like `PersonsResponsePage` so the purpose is a bit clearer? (same for others)

SayTheirNames/Source/Service/Networking/NetworkRequestor.swift
```diff
@@ -0,0 +1,44 @@
...
+    let service: Service
+    let concurrentQueue = DispatchQueue(label: "NetworkRequestor", attributes: .concurrent)
+        
+    required init(service: Service) {
+        self.service = service
+    }
+    
+    // MARK: - Public methods
+    
+    public func fetchDecodable<T: Decodable>(_ url: String, completion: @escaping (T?) -> Swift.Void) {
```
Kilo-Loco, line 24: I think it would provide more flexibility and allow us to handle errors elegantly if we used `Result` instead
```suggestion
    public func fetchDecodable<T: Decodable>(_ url: String, completion: @escaping (Result<T, Error>) -> Swift.Void) {
```

SayTheirNames/Source/Service/Networking/NetworkRequestor.swift
```diff
@@ -0,0 +1,44 @@
...
+        let request = AF.request(url)
+        request.responseDecodable(of: T.self, queue: self.concurrentQueue) { (response) in
+            DispatchQueue.mainAsync { completion(response.value) }
+        }
+    }
+    
+    public func fetchData(_ url: String, completion: @escaping (Data?) -> Swift.Void) {
```
Kilo-Loco, line 31:
```suggestion
    public func fetchData(_ url: String, completion: @escaping (Result<Data, Error>) -> Swift.Void) {
```

SayTheirNames/Source/Service/Networking/BaseNetworkUrl.swift
```diff
@@ -0,0 +1,24 @@
...
+            return "https://saytheirnames.dev"
+        #elseif QA
+            return "https://saytheirnames.qa"
+        #elseif LOCAL
+            return "https://localhost"
+        #else
+            return "https://saytheirnames.dev" // switch to PROD
+        #endif
+    }
+}
```
hybridcattt, line 24: This could be simplified to smth like:
```
enum Environment {
    static let serverURLString = {
        #if ... 
    }()
}
```
and used like `Environment.serverURLString`

SayTheirNames/Source/Service/Networking/NetworkRequestor.swift
```diff
@@ -0,0 +1,44 @@
...
+        let request = AF.request(url)
+        request.responseData(queue: self.concurrentQueue) { (response) in
+            DispatchQueue.mainAsync { completion(response.value) }
+        }
+    }
+    
+    public func fetchJson(_ url: String, completion: @escaping (Any?) -> Swift.Void) {
```
Kilo-Loco, line 38: I think AF has a `JSON` type that we might be able to use here but i could be wrong
```suggestion
    public func fetchJson(_ url: String, completion: @escaping (Result<JSON, Error>) -> Swift.Void) {
```

SayTheirNames/Source/Service/Networking/Requestors/DonationsNetworkRequestor.swift
```diff
@@ -0,0 +1,24 @@
...
+
+import Foundation
+
+// MARK: - PetitionUrl
+final class DonationsUrl: BaseNetworkUrl {
+    class var donations: UrlString { return "\(self.base)/api/donations" }
+}
+
+// MARK: - DonationsNetworkRequestor
+typealias DonationsNetworkRequestor = NetworkRequestor
```
hybridcattt, line 17: I honestly think that these type aliases make it more confusing, as one would need to jump one more definition to see what it is. So just an extension on `serverURLString` would be clearer IMO.

SayTheirNames/Source/Service/Networking/Requestors/DonationsNetworkRequestor.swift
```diff
@@ -0,0 +1,24 @@
...
+extension DonationsNetworkRequestor {
+    // MARK: - Public methods
+    
+    public func fetchDonations(completion: @escaping (Donations?) -> Swift.Void) {
```
hybridcattt, line 21: Let's make completion have `Result` already now, so we can debug errors more easily

SayTheirNames/Source/Service/Networking/NetworkRequestor.swift
evilpenguin: Adding

SayTheirNames/Source/Service/Networking/NetworkRequestor.swift
evilpenguin: Adding

SayTheirNames/Source/Service/Networking/NetworkRequestor.swift
evilpenguin: Adding

SayTheirNames/Source/Controller/Home/HomeController.swift
evilpenguin: 🔥

SayTheirNames/Source/Model/Donation.swift
evilpenguin: Fixed.

SayTheirNames/Source/Model/Person.swift
evilpenguin: I can make it empty - it was choice.

SayTheirNames/Source/Service/Networking/Requestors/DonationsNetworkRequestor.swift
evilpenguin: Indeed.

SayTheirNames/Source/Model/Person.swift
hybridcattt: my thinking is that if there's no meaningful difference between nil and [] (i.e. both mean that there's no media) - then it's easier to handle at usage site

evilpenguin: I think handling `nil` at the UI level would be best. This way we don't have floating instance and /wasting/ memory for a simple check.

SayTheirNames/Source/Model/Image.swift
evilpenguin: Fixed.

SayTheirNames/Source/Service/Networking/NetworkRequestor.swift
evilpenguin: Fixed

SayTheirNames/Source/Model/Person.swift
hybridcattt: at the usage site you'd have to do
```
if person?.socialMedia.isEmpty == false ?? false {
   // hide
}
```
or
```
if let sm = person?.socialMedia, !sm.isEmpty {
   // hide
}
```
because we don't want to show the section if it's an empty array. 

With non-optional check it's simply an emptiness check. Empty array almost doesn't take up any memory, and nil is a case of an `Optional` enum which also takes memory.

evilpenguin: The backend will not provide an empty array for a given key - thus at parsing time it will not return an empty array.

That said the check would simply be:

`if let sm = person?.socialMedia {
   // hide/show section
}`

I agree but the memory use for an `Array` vs `nil` is a vast difference.

SayTheirNames/Source/Model/Person.swift
evilpenguin: I like the name. Changed for all.

SayTheirNames/Source/Service/Networking/BaseNetworkUrl.swift
evilpenguin: Fixed.

SayTheirNames/Source/Service/Networking/Requestors/DonationsNetworkRequestor.swift
evilpenguin: I would disagree, but I will gladly change it. Changed.

SayTheirNames/Source/Model/Person.swift
hybridcattt: I would agree, if we could rely on the backend always sending non-empty arrays. On the client side we have to account for all cases though - if there comes null, wrong type, etc - the app should work gracefully and not crash / show buggy UI.

We can still check it as-is, just more characters. Up to you, anyway, it's a minor thing 😊

evilpenguin: I added a `stud` for each, so we no longer have `Optional`s.

ohwhen:
I suggest to split this PR up in multiple PR's...make one for the AlamoFire dependency and one for the networking layer. Easier to review and comment.

SayTheirNames/Source/Controller/Home/HomeController.swift
rodericj: Is the fact that this is a `network` really relevant to the HomeController? Maybe a bit pedantic but it feels like it's not really what we want to convey to the controller.

SayTheirNames/Source/Service/Service.swift
```diff
@@ -13,7 +13,8 @@ final class Service {
     lazy private(set) var navigator = Navigator(service: self)
```
rodericj, line 13: Related to my comment about the network being added in this class, this whole class feels like it could be a protocol. Probably out of scope of this PR, but it would really help us in testing these controllers with a `Servicing` protocol that defines `navigator`, `image`, `dateFormatter`, and `dataStore` (rather than `network`). The `Service` class could implement the protocol, and we could just pass around something that conforms to the protocol. 


Ok this is definitely out of scope for this PR.

evilpenguin: Completely agree. Sounds like a solid improvement for another PR. Haha

PR #59, opened 2020-06-04
Add image loading library

juanfolivera:
**Closed the previous PR due to too many conflicts**

Addressed comments from the [previous PR](https://github.com/Say-Their-Name/say-their-names-ios/pull/50)

ezgif com-video-to-gif


SayTheirNames/Source/Utils/Images/STNImage.swift
```diff
@@ -0,0 +1,20 @@
...
+//  SayTheirNames
+//
+//  Created by Juan Olivera on 6/4/20.
+//  Copyright © 2020 Franck-Stephane Ndame Mpouli. All rights reserved.
+//
+
+import Foundation
+import Kingfisher
+
+struct STNImage {
```
hybridcattt, line 12: `enum` is more common for namespacing as it doesn't have an `init()`

hybridcattt: what do you think about having an extension on `UIImageView` instead?

juanfolivera: I ended up doing an extension now it will be just setting an image with: 
`personImageView.populate(withURL: url)`

SayTheirNames/Source/View/Home/Person/PersonCell.swift
```diff
@@ -77,9 +77,10 @@ final class PersonCell: UICollectionViewCell {
     }
     
     func configure(with person: Person) {
-        profileImageView.image = UIImage(named: "man-in-red-jacket-1681010")
-        // ^Update to default place holder image when resource is added.
-        // TO-DO: Use Kilo's ImageDownloader here to DL image
+        let df = DateFormatter()
+        df.dateFormat = "dd.MM.yyyy"
+        
+        profileImageView.populate(withURL: person.images.first?.personURL ?? "")
```
evilpenguin, line 83: If we add the following to `Image.swift` we can remove the nil-coalescing operator here.

```
    public init(from decoder: Decoder) throws {
        let values = try decoder.container(keyedBy: CodingKeys.self)
        
        self.id = try values.decodeIfPresent(Int.self, forKey: .id) ?? -1
        self.personID = try values.decodeIfPresent(String.self, forKey: .personID) ?? ""
        self.personURL = try values.decodeIfPresent(String.self, forKey: .personURL) ?? ""
    } 
```

This way the model will always return some value for the image.

SayTheirNames/Source/View/Home/Person/PersonPhotoCell.swift
```diff
@@ -35,10 +35,11 @@ final class PersonPhotoCell: UICollectionViewCell {
     private func setup() {
         addSubview(personImageView)
         personImageView.fillSuperview()
+        personImageView.clipsToBounds = true
```
evilpenguin, line 38: Lets move this to the `personImageView` computed property.

SayTheirNames/Source/View/Home/Person/PersonCell.swift
juanfolivera: I'm not sure I follow, the ticket originally said to add a well-mantained library for images. 

Attaching card:
https://trello.com/c/GBTondmf/37-add-an-image-loading-library

juanfolivera:
Had to remove Kingfisher cause it was causing swiftlint errors, I changed it for [sdWebImage](https://github.com/SDWebImage/SDWebImage) which is pretty reliable too

SayTheirNames/Source/View/Home/Person/PersonCell.swift
evilpenguin: Lets skip this for now. We can do it in another PR. Thank you.

PR #61, opened 2020-06-04
Add Servicing protocol to define the behavior of the service

rodericj:
I've added a servicing protocol to define the behavior of the `Service` object. This will help out with testing the consumers of this service in a followup PR


rodericj:
I can't edit labels, but I did rebase so it's no longer conflicting.

PR #67, opened 2020-06-05
[CHORE] Added Spanish Localization

PedroTrujilloV:
Added Spanish localization for: PersonHeaderCell, PersonController, also added Localization for new static text for English, Spanish and Russian. 
@Say-Their-Name/ios


PedroTrujilloV:
@hybridcattt I'm not able to add reviewers

SayTheirNames/Source/Controller/Home/Person/PersonController.swift
```diff
@@ -9,14 +9,36 @@
...
+            self.readMoreButton.setTitle("READ MORE ->", for: UIControl.State.selected)
+        }
+        
+        if let donate = NSLocalizedString("Donate", comment: "Donate") as String? {
+            self.donateButton.setTitle(donate, for: UIControl.State.normal)
+            self.donateButton.setTitle(donate, for: UIControl.State.selected)
+        } else {
+            self.donateButton.setTitle("Donate", for: UIControl.State.normal)
+            self.donateButton.setTitle("Donate", for: UIControl.State.selected)
+        }
```
Sharkesm, line 38: @PedroTrujilloV Is there a better way to include Localized strings without having `NSLocalizedString` to keep things neat??

PedroTrujilloV: I can add it as an array `[.normal, .selected]` because currently the string is the same, to avoid some duplication. Is it what you mean?

jaywardell:
Just know that this is a moving target.  https://github.com/Say-Their-Name/say-their-names-ios/pull/69 added another key to Localizable.strings.  This will be changing a lot for a while.

SayTheirNames/Source/Controller/Home/Person/PersonController.swift
rodericj: I sort of thought we were using https://github.com/SwiftGen/SwiftGen for localizations. If so these should be something like `L10n.donate`, right? I'm not sure if it's documented anywhere, but I do see [this](https://github.com/Say-Their-Name/say-their-names-ios/blob/feature/home_diffable_data_source/SayTheirNames/Resources/Generated/strings.swift) generated file of localizations.

Sharkesm: @rodericj Yes, and thanks for the reference 💯

hybridcattt: good idea to document it 💯

SayTheirNames/Resources/Base.lproj/Localizable.strings
```diff
@@ -3,13 +3,26 @@
   SayTheirNames
 
   Created by Kyle Lee on 6/1/20.
+  Edited by Pedro Trujillo on 6/4/20
   Copyright © 2020 Franck-Stephane Ndame Mpouli. All rights reserved.
 */
 
+/* Tab Bar */
```
Kilo-Loco, line 10: I agree that having the Strings separated by comments makes everything look more organized, but the problem is that we may be reusing these strings in other places. Cancel for example may end up on several screens but we are currently only using it in search. This then leads to the problem of moving strings around when they need to be added somewhere else.

IMO, there are only two ways to approach the organization of these strings that will allow the app to scale:
1. We continue to separate strings into sections, but use dot syntax so the generated strings can be grouped together when accessing them.
Example:
```
/* Search */
"search.cancel" = "Cancel";

/* Home */
"home.cancel" = "Fancy Cancel";
```
```
label.text = Strings.Search.cancel
label.text = Strings.Home.cancel
```

2. We keep all the strings alphabetized by their key value and simply add more descriptive keys for different use case.
Example:
```
"cancel" = "Cancel";
"cancel_fancy" = "Fancy Cancel" // we could also do "fancy_cancel"
```
```
label.text = Strings.cancel
label.text = Strings.cancelFancy
```

evilpenguin:
Whats the status of this PR?

Kilo-Loco:
Closing due to inactivity. Please push an updated branch or reach out on Slack if you would like to reopen

PR #69, opened 2020-06-05
All programmatic UI views now support dynamic type

jaywardell:
All programmatic UI views now support dynamic type
view layout could be better at very large sizes, but at least things work


SayTheirNames/Source/View/Home/Person/PersonCell.swift
```diff
@@ -108,6 +110,26 @@ final class PersonCell: UICollectionViewCell {
...
+    private func styleLabels() {
+
+        nameLabel.font = UIFont.STNDynamic.cardTitle
+        dateOfIncidentLabel.font = UIFont.STNDynamic.cardSubitle
+    }
+    
+    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        super.traitCollectionDidChange(previousTraitCollection)
+        
+        if traitCollection.preferredContentSizeCategory != previousTraitCollection?.preferredContentSizeCategory {
```
hybridcattt, line 128: feels like we don't need this bit 🤔  

because fonts in our extension are not self-resizing by intention (then UIs don't need to be so dynamic). users usually don't change text settings on the fly, only developers do :) 
if we wanted fonts to self-resize, there's a way to create fonts that change by themselves without us writing extra code to react to change (using font metrics)

SayTheirNames/Source/View/Home/Person/PersonCell.swift
```diff
@@ -23,7 +23,7 @@ final class PersonCell: UICollectionViewCell {
     private lazy var nameLabel: UILabel = {
         let lbl = UILabel()
         lbl.text = "GEORGE FLOYD"
-        lbl.font = UIFont(name: "Karla-Bold", size: 13)
+//        lbl.font = UIFont(name: "Karla-Bold", size: 13)
```
Kilo-Loco, line 26:
```suggestion
```

SayTheirNames/Source/View/Home/Person/PersonCell.swift
```diff
@@ -33,7 +33,7 @@ final class PersonCell: UICollectionViewCell {
     private lazy var dateOfIncidentLabel: UILabel = {
         let lbl = UILabel()
         lbl.text = "25.05.2020"
-        lbl.font = UIFont(name: "Karla-regular", size: 13)
+//        lbl.font = UIFont(name: "Karla-regular", size: 13)
```
Kilo-Loco, line 36:
```suggestion
```

SayTheirNames/Source/View/Home/Person/PersonCell.swift
jaywardell: sorry for the dead code. fixed

SayTheirNames/Source/View/Home/Person/PersonCell.swift
jaywardell: sorry for the dead code. fixed

SayTheirNames/Source/View/Home/Person/PersonCell.swift
jaywardell: If I take out the call from traitCollectionDidChange and test on device, then the text size is not updated until the app is restarted.  For most users, that means waiting until either 1) the app is quit in the background after they've not used it for a while or 2) the phone is restarted.  Tech-save people force-quit their apps.  Most typical users don't, or wouldn't necessarily think of it in the moment.

I'm thinking of 2 scenarios in particular:

1) Someone downloads STN for the first time and launches it to discover that the text size is too small for them.  They go to the system preferences and try to increase the type size.  They return to STN and see that nothing has changed.  They immediately decide the app isn't that useful if they have to squint to read it.

2) Two friends are talking about BLM in the news.  One remembers reading something in STN and loads the story and hands her phone to the other, who is a slightly older person with poorer eyesight. The second friend asks the first to make the screen more readable.  The first tries, but the text size doesn't change. Rather than try to troubleshoot on the spot, they give up.

I see this as an accessibility issue and a usability issue.  I'd politely ask that you reconsider.

SayTheirNames/Source/View/Home/Person/PersonCell.swift
jaywardell: please see my response to hybridcatttre: trait changes

SayTheirNames/Source/View/Home/Person/PersonCell.swift
Kilo-Loco: It definitely couldn't hurt to have an app be more accessible, but i think anyone that is familiar with quickly switching font sizes on iphone would also be familiar with killing apps as well. 

Overall, what i would prefer to see here is that we replace the current STN font code with the computed code so we can refactor in place.

jaywardell: >> Overall, what i would prefer to see here is that we replace the current STN font code with the computed code so we can refactor in place.
 
Are you saying replace the properties in STN with the ones in STNDynamic?  I'd definitely like to get there eventually, but didn't want to break other people's code for now with the rate that PRs are coming in. I think this branch is already about 20 commits behind...

>> ...but i think anyone that is familiar with quickly switching font sizes on iphone would also be familiar with killing apps as well.

When's the last time you helped a teacher over 50 with their phone? They tend to see these kinds of features in isolation, not necessarily understand how to use them in combination.  Not to pick on educators, I was one myself until last year...

Kilo-Loco: Lets remove `STNDynamic` and refactor the code in `STN` to be dynamic
```
static let title = UIFont.dynamicCustomFont(
        fontName: FontName.karlaBold.rawValue,
        fontName: FontName.karlaBold.rawValue,
        textStyle: .title1)
        textStyle: .title1)
```
to
```
static var title: UIFont { UIFont.dynamicCustomFont(
        fontName: FontName.karlaBold.rawValue,
        textStyle: .title1) }
```

jaywardell: easy enough. done

jaywardell:
--updated to include PersonController

jaywardell:
rebased again after another big dump of changes in development

jaywardell:
I'm going to visit my parents for the first time in 3 months, so I'll be mostly afk for the next 24 hours.  I'd love for this to be merged into development before I have to rebase again. If you have a minor issue with this PR, please dm me, or better yet add a Trello card AND dm me.

jaywardell:
...thanks

hybridcattt:
thanks a lot @jaywardell 🙌

PR #70, opened 2020-06-05
Added FR language for L10N.

npvisual:
Basic translation of the current localized strings.
Might need to be revisited when there is more data.

Also PR #67 is reorganizing the L10n base file so we might want to either wait to push this one or have another PR once #67 is merged.


rodericj:
y'all are impressive with these localizations.

PR #73, opened 2020-06-05
[DELETE] Removed PersonDetailsController with associated xib

Sharkesm:
Note: It was confirmed by the design team that we no longer need `PersonDetails` screen


dasautoooo:
Think this can be done in #75 . It doesn't matter if delete those files now or later alone with #75 being merged.
Maybe we close it?

evilpenguin:
Are we closing this?

dasautoooo:
The changes are already in #88. 
Closed it.

Sharkesm:
@dasautoooo Great catch :))

PR #76, opened 2020-06-05
[CHORE] Set date formatter to respect local date and time format.

Isuru-Nanayakkara:
- Changed locale and timeZone property values in date formatters to current.
- Added extension method to show localized strings from dates.


Say Their Names/Source/Service/DateFormatter/DateFormatterService.swift
```diff
@@ -29,7 +29,8 @@ class DateFormatterService {
             }
             
             let dateFormatter = DateFormatter()
-            dateFormatter.locale = Locale(identifier: "en_US_POSIX")
+            dateFormatter.locale = .current
+            dateFormatter.timeZone = .current
```
hybridcattt, line 33: aren't these the default values?

Say Their Names/Source/View/Home/Person/PersonCell.swift
```diff
@@ -77,14 +77,11 @@ final class PersonCell: UICollectionViewCell {
...
     func configure(with person: Person) {
-        let df = DateFormatter()
-        df.dateFormat = "dd.MM.yyyy"
-        
         profileImageView.image = UIImage(named: "man-in-red-jacket-1681010")
         // ^Update to default place holder image when resource is added.
         // TO-DO: Use Kilo's ImageDownloader here to DL image
         nameLabel.text = person.fullName.uppercased()
-        dateOfIncidentLabel.text = df.string(from: person.date) 
+        dateOfIncidentLabel.text = person.date.localizedString(dateStyle: .short)
```
hybridcattt, line 84: The idea is that all date formatters are cached in `DateFormatterService` and reused. I see this one wasn't migrated to that yet, but now could be a good time to do it? Or can create a to-do ticket on Trello. Up to you :)

Say Their Names.xcodeproj/project.pbxproj
```diff
@@ -1203,6 +1207,7 @@
 			isa = XCBuildConfiguration;
 			buildSettings = {
 				CODE_SIGN_STYLE = Automatic;
+				DEVELOPMENT_TEAM = H4TYCSEM4B;
```
hybridcattt, line 1210: please revert these 🙏

Isuru-Nanayakkara: Sorry about these. I forgot to clear them up before submitting. WIll remove those.

Say Their Names/Source/View/Home/Person/PersonCell.swift
Isuru-Nanayakkara: The current implementation of caching formatters is done by using the given format as the key, yes? But the `DateFormatter.localizedString` method is a class function so I can use it without creating an instance of `DateFormatter` and passing a format. So I'm wondering how to fit that into the caching mechanism.

Say Their Names/Source/Service/DateFormatter/DateFormatterService.swift
Isuru-Nanayakkara: Yes, you're right. I removed those. 👍🏼

Say Their Names/Source/View/Home/Person/PersonCell.swift
hybridcattt: yeah it's a bit confusing. I'd say there's no need to cache by key as it is done now - we will have only a handful of dates, so they can be just individual variables like `dateOfIncidentFormatter` etc each with their own configuration. 

Having a function like `localizedString...` is another approach which is also viable, though in this project date/time formatting has been agreed to be done via `DateFormatterService`, and it's good to stick to one way of doing it within a project 😊

Say Their Names.xcodeproj/project.pbxproj
hybridcattt: No worries! it's a common thing, happens to many here :)

Isuru-Nanayakkara:
- Deleted Date extension file.
- Moved localised date formatter method to DateFormatterService.

Say Their Names/Source/View/Home/Person/PersonCell.swift
Isuru-Nanayakkara: I agree. I moved the date formatting code to the `DateFormatterService`. Btw I was referring to the `[localizedString](https://developer.apple.com/documentation/foundation/dateformatter/1415241-localizedstring)` function in iOS itself earlier. I had also named my extension method the same name so that must have been confusing. Anyway, no more extension methods now.

Say Their Names/Source/View/Home/Person/PersonCell.swift
```diff
@@ -77,14 +77,11 @@ final class PersonCell: UICollectionViewCell {
...
     func configure(with person: Person) {
-        let df = DateFormatter()
-        df.dateFormat = "dd.MM.yyyy"
-        
         profileImageView.image = UIImage(named: "man-in-red-jacket-1681010")
         // ^Update to default place holder image when resource is added.
         // TO-DO: Use Kilo's ImageDownloader here to DL image
         nameLabel.text = person.fullName.uppercased()
-        dateOfIncidentLabel.text = df.string(from: person.date) 
+        dateOfIncidentLabel.text = DateFormatterService().localizedString(date: person.date, dateStyle: .short)
```
Mackis, line 84: @hybridcattt what about making services singletons. 

`DateFormatterService.shared.localizedString(date: person.date, dateStyle: .short)` 

So we are not creating instances of services in multiple places. The dateFormatter is only being used here for now but could potentially grow.

hybridcattt: Very good point about singleton, just I think it should rather be
`DateFormatterService.shared.dateOfIncidentString(date)` , because the goal is to have consistent date styles across the app while keeping the code duplication low, and not so much that default API is not enough. 
The internal implementation of that kind of method can even use `localizedString...` indeed like you mentioned above 👍

kingsqd: @hybridcattt @Mackis when first implementing the DateFormatter service I created it as a singleton, but got some push back that we should use it in Service.  I agree with @Kilo-Loco  that we probably should be passing it in as an arg to PersonCell.

evilpenguin: Interesting. The `Person` model doesn't have a `date` object? Where is this coming from?

evilpenguin: We are also going with dependency injection, so we'd need our cells and whatnot to work with `ServiceReferring`.

evilpenguin:
It looks like this branch is using out dated code? The models are many many revisions ago.

hybridcattt:
@evilpenguin in the latest model `doi` is listed as `String` which is not going to work if we want to localize the date format. Will create a to-do to fix

evilpenguin:
Whats going on with this PR? It's 3 days old now.

hybridcattt:
I started looking into this, will make a revamp without conflicts shortly

hybridcattt:
See #76

hybridcattt:
Thanks for having a look @Isuru-Nanayakkara !

PR #80, opened 2020-06-05
[Chore] Opt-in to trigger Object Literals as warnings in SwiftLint

Kilo-Loco:
We should not use image literals (`#imageLiteral`) in the project since it can cause the app to crash if the asset get replaced or renamed. SwiftLint groups colors into this category as well so the same rules apply


jaywardell:
definitely agree with this

PR #81, opened 2020-06-05
Adding MIT license to project

ohwhen:
Clarify the project is being published as opensource, making use of one of the more permissive non-copyleft licenses possible, the MIT license. 

I've used the same format as projects like Alamofire and SnapKit are using, a LICENSE file in the root and a license header at the start of each file.

See for more info here: https://choosealicense.com/licenses/mit/


evilpenguin:
@ohwhen Why did you delete the branch?

PR #85, opened 2020-06-05
[Chore] Add theme file

bertadevant:
Theme file for UI constants split into components vs screens. 

First half of the work, trying to avoid making a huge PR


SayTheirNames/Source/Utils/Theme/Theme.swift
```diff
@@ -0,0 +1,62 @@
...
+        // MARK: - Button
+        enum Button {
+            enum Size {
+                static let medium: CGSize = .init(width: 40, height: 40)
+            }
+        }
+    }
+    
+    // MARK: - Screens
+    enum Screens {
```
Mackis, line 32: Nicely done! 

Don't think we should separate styling per screen. We could drop `enum home` & `enum screens` this way we have standard sizes across screens. Should we need custom per screen we can do `static let homePeopleHeight = 170`

bertadevant: I was thinking the same way but current screens really do use different sizes and I am not sure if I should be the one to standardize things, so i wanted to have all put together and then on a separate PR talk to design and get standards!

hybridcattt: Sounds like a perfect case for a follow-up to-do indeed - that would be a nice small ticket to work on

evilpenguin:
@bertadevant Can you fix the conflicts?

bertadevant:
@evilpenguin yes currently working on it, rebasing is new to me is taking a bit longer than usual 😅

SayTheirNames/Source/Controller/Home/Person/PersonDetailsController.swift
```diff
@@ -0,0 +1,57 @@
+//
+//  PersonInfoViewController.swift
+//  SayTheirNames
+//
+//  Created by Manase on 31/05/2020.
+//  Copyright © 2020 Franck-Stephane Ndame Mpouli. All rights reserved.
+//
```
evilpenguin, line 7: Can we get the license here?

SayTheirNames/Source/Utils/Extentions/UIEdgeInsets+init.swift
```diff
@@ -0,0 +1,24 @@
+//
+//  UIEdgeInsets+init.swift
+//  SayTheirNames
+//
+//  Created by Berta Devant on 05/06/2020.
+//  Copyright © 2020 Franck-Stephane Ndame Mpouli. All rights reserved.
+//
```
evilpenguin, line 7: License here as well.

SayTheirNames/Source/Utils/Theme/Theme.swift
```diff
@@ -0,0 +1,62 @@
+//
+//  Theme.swift
+//  SayTheirNames
+//
+//  Created by Berta Devant on 05/06/2020.
+//  Copyright © 2020 Franck-Stephane Ndame Mpouli. All rights reserved.
+//
+
```
evilpenguin, line 8: License here as well.

Sharkesm:
@bertadevant Great work ⭐

PR #90, opened 2020-06-06
Added more network requestors

evilpenguin:
### Added
- Full search
- Person search (name, city, country)
- Petition search (name, type)
- Donation search (name, type)

### DIFF
--- a/SayTheirNames/Source/Model/Search.swift
+++ b/SayTheirNames/Source/Model/Search.swift
@@ -1,5 +1,5 @@
 //
-//  Media.swift
+//  Petition.swift
 //  SayTheirNames
 //
 //  Copyright (c) 2020 Say Their Names Team (https://github.com/Say-Their-Name)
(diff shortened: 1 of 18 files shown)


SayTheirNames/Source/Service/Service.swift
```diff
@@ -42,5 +42,14 @@ final class Service: Servicing {
         Log.mode = .all
         Log.print("STN Version: \(Bundle.versionBuildString)")
         Log.print("Starting Services")
+        
+        self.network.fetchPeopleByName("george") { (result) in
```
Kilo-Loco: should we remove this?

evilpenguin: Yeah I pushed after you started the review.

SayTheirNames/Source/Model/Search.swift
```diff
@@ -0,0 +1,46 @@
...
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+
+import Foundation
+
+public struct SearchResponsePage: Decodable {
+    let people: [Person]
+    let donations: [Donation]
+    let petitions: [Petition]
+    let link: Link
```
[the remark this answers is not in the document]
evilpenguin: We will always have a `Link` object from the backend. The values will just be `nil` some cases.

evilpenguin:
> Kickstarting workflow again...

Couple changes coming!

evilpenguin:
> I'm wondering if we should be working with `URLQueryItem`s instead of using String interpolation to construct the urls

Added

PR #91, opened 2020-06-06
Add voice over support

Eugeneberezin:
What this PR do?
I added accessibility labels and enabled voice-over feature for all of the UIs. 
Some of the accessibility labels have to be spelled phonetically so VO can read it correctly. 

How to test? 
On a device: enable voice over and open the app. Try to go through each UI. (try not to get lost)
On a simulator: open accessibility inspector and enable voice over. It will automatically go through the UIs and read them out.

### DIFF
--- a/SayTheirNames/Source/Controller/Home/CollectionViewDataSource/PersonCollectionViewDataSource.swift
+++ b/SayTheirNames/Source/Controller/Home/CollectionViewDataSource/PersonCollectionViewDataSource.swift
@@ -45,11 +45,14 @@ final class PersonCollectionViewDataSourceHelper {
             cell.configure(with: person)
             cell.accessibilityIdentifier = "peopleCell\(indexPath.item)"
             cell.isAccessibilityElement = true
+            cell.accessibilityNavigationStyle = .automatic
+            cell.accessibilityLabel = "\(person.fullName), tap to learn more"
             return cell
         }
         
         self.dataSource.supplementaryViewProvider = { (collectionView, kind, indexPath) in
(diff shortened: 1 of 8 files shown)


evilpenguin:
I feel like we should have some some /helper/ object for all of this? Maybe even use SwiftGen or lend a hand?

Eugeneberezin:
@evilpenguin voice-over and accessibility can be a bit tricky. I personally don't really know a good way to create a general helper that would fit all of the cases.

SayTheirNames/Source/Controller/Home/CollectionViewDataSource/PersonCollectionViewDataSource.swift
```diff
@@ -45,11 +45,14 @@ final class PersonCollectionViewDataSourceHelper {
             cell.configure(with: person)
             cell.accessibilityIdentifier = "peopleCell\(indexPath.item)"
             cell.isAccessibilityElement = true
+            cell.accessibilityNavigationStyle = .automatic
+            cell.accessibilityLabel = "\(person.fullName), tap to learn more"
```
VidaHasan, line 49: Can you remove "tap to learn more" from accessibility label
I think HIG recommends not to tell users how to interact with label manually. Instead put "Learn more" in accessibility hint.
https://developer.apple.com/documentation/objectivec/nsobject/1615093-accessibilityhint#:~:text=An%20accessibility%20hint%20helps%20users,obvious%20from%20the%20accessibility%20label.

Eugeneberezin: Good point! Done

PR #97, opened 2020-06-06
[Chore] Remove BaseViewController and implement ServiceReferring

emelagumat:
- BaseViewController has been removed and replaced with UIViewController+ServiceReferring on Controllers which were inheriting from it.

😅👇
- If possible, I would appreciate if someone can explain me why this option is better than making BaseViewController implements ServiceReferring and make all controllers inherit from it.


Sharkesm:
@evilpenguin @Kilo-Loco Care to explain on this :)

Mackis:
One of the main reasons Protocols are preferred over subclassing is because Swift if a Protocol Oriented Language. 

Why is `BaseViewController` frowned upon? 
`BaseViewController` is viewController which defines behaviour for it's child classes. This is something we want to avoid. We would like our classes to define their own behaviour. 

Which is why we prefer protocols. They define properties and functionality but do not dictate to their child classes how to behave. 
Yes this can be achieved by subclassing but for example, a bad change in `BaseViewController`s viewDidLoad method could lead to problems in all child viewControllers. 

Protocols = Blueprint 
Subclassing = Instructions 

This is a good practical example https://talk.objc.io/episodes/S01E29-protocols-class-hierarchies

Hope this helps

emelagumat:
> One of the main reasons Protocols are preferred over subclassing is because Swift if a Protocol Oriented Language.
> 
> Why is `BaseViewController` frowned upon?
> `BaseViewController` is viewController which defines behaviour for it's child classes. This is something we want to avoid. We would like our classes to define their own behaviour.
> 
> Which is why we prefer protocols. They define properties and functionality but do not dictate to their child classes how to behave.
> Yes this can be achieved by subclassing but for example, a bad change in `BaseViewController`s viewDidLoad method could lead to problems in all child viewControllers.
> 
> Protocols = Blueprint
> Subclassing = Instructions
> 
> This is a good practical example https://talk.objc.io/episodes/S01E29-protocols-class-hierarchies
> 
> Hope this helps

Yes, it helped! Thanks!! 🙌😁

evilpenguin:
👏

PR #44, opened 2020-06-04
[UI] Added app icon

NicolasR27:
just added the app icon png from resource  folders.

PR #105, opened 2020-06-06
[CHORE] added icon

evilpenguin:
Added the app icons


hybridcattt:
I see the original PR #44 also had image files included, they appear to be missing here

evilpenguin:
> I see the original PR #44 also had image files included, they appear to be missing here

Pushed

evilpenguin:
Being worked on: the icons were wrong from #44. We are also missing 57/57@2x

PR #107, opened 2020-06-06
Implement donations controller

Kilo-Loco:
This PR implements almost everything needed for the Donations Controller
- The filters at the top have been implemented in a reusable way
- The donations collection view was built with a manager that uses diffable data source in a reusable way
- Many other miscellaneous abstractions that allow cleaner, more readable code

[
Screen Shot 2020-06-06 at 4 11 29 PM
](url)


SayTheirNames/Source/Utils/Extentions/UIStackView++.swift
```diff
@@ -0,0 +1,29 @@
...
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+
+import UIKit
+
+extension UIStackView {
+    func addArrangedSubviews(_ views: UIView...) {
+        views.forEach { self.addArrangedSubview($0) }
+    }
+}
```
dillon-mce, line 29: 🎉

SayTheirNames/Source/Utils/Color/UIColor+Resource.swift
```diff
@@ -26,6 +26,12 @@ import UIKit
 
 extension UIColor {
     
+    private static func dynamic(light: UIColor, dark: UIColor) -> UIColor {
+        return .init { trait in
+            return trait.isDarkModeOn ? dark : light
+        }
+    }
+
```
dillon-mce, line 34: I don't know that this is clearer than just using the initializer that you're wrapping. It is also kind of limited. What if you wanted to add colors for high contrast?

SayTheirNames/Source/View/Filter/FilterCategoryCell.swift
```diff
@@ -24,60 +24,61 @@
...
-class LocationCell: UICollectionViewCell {
+final class FilterCategoryCell: UICollectionViewCell {
     
-    var titleLabel: UILabel = {
-        let label = UILabel()
-        label.textAlignment = .center
-        return label
-    }()
+    private var titleLabel = UILabel.create {
+        $0.translatesAutoresizingMaskIntoConstraints = false
```
dillon-mce, line 30: This line and lines 61-66 were all accomplished in `titleLabel.fillSuperView()` (L50) before. Is the style you have here preferred? I assumed we should be using those convenience methods.

SayTheirNames/Source/Utils/Color/UIColor+Resource.swift
Kilo-Loco: I added this so we could at least start adding dynamic colors to the app so they work during dark mode. definitely a lot of room to improve here in a different PR

SayTheirNames/Source/View/Filter/FilterCategoryCell.swift
Kilo-Loco: Did this to be consistent with the rest of the code in the diff but will add a quick follow up diff to use `fillSuperView()`

PR #113, opened 2020-06-07
[DRAFT] Update Service to Dependency (Using properyWrapper)

evilpenguin:
Looking for feedback on this. Thoughts?


SayTheirNames/Source/View/Home/Person/PersonCell.swift
```diff
@@ -103,7 +103,7 @@ final class PersonCell: UICollectionViewCell {
     func configure(with person: Person) {
         profileImageView.populate(withURL: person.images.first?.personURL ?? "")
         nameLabel.text = person.fullName.uppercased()
-        dateOfIncidentLabel.text = person.doi
+        dateOfIncidentLabel.text = person.doi.description
```
evilpenguin, line 106: Please ignore this, haha.

SayTheirNames/Source/Dependency/DependencyContainer.swift
```diff
@@ -24,23 +24,63 @@
...
+// MARK: - Dependency
+protocol Dependency {
+    
 }
-/// This is a core class that holds all instances responsible for logic
-final class Service: Servicing {
+
+// MARK: - DependencyContainer
+final class DependencyContainer: Dependency {
+    lazy private(set) var navigator = Navigator()
```
hybridcattt, line 34: can it be `let` ? the deps should be lightweight to create so I think it's ok if not lazy

SayTheirNames/Source/Dependency/Networking/NetworkRequestor.swift
```diff
@@ -27,32 +27,33 @@ import Alamofire
 
 // MARK: - NetworkRequestor
 
-final class NetworkRequestor {
-    let concurrentQueue = DispatchQueue(label: "NetworkRequestor", attributes: .concurrent)
+final class NetworkRequestor: Dependency {
     let session: Session
-
+    let concurrentQueue = DispatchQueue(label: "NetworkRequestor", attributes: .concurrent)
+    private let headers = HTTPHeaders.init(["x-dates-epoch": "true"])
```
hybridcattt, line 33: could you make a separate PR for the date stuff, for visibility? 🙏

SayTheirNames/Source/Dependency/DependencyContainer.swift
```diff
@@ -24,23 +24,63 @@
...
-    static let shared = Service()
-    
     // MARK: - Init
     init() {
         Log.mode = .all
         Log.print("SayTheirNames Version: \(Bundle.versionBuildString)")
         Log.print("Starting Services")
+        
+        // Handle our Injection
+        let injectionFactory = DependencyInjectionFactory.notShared
```
hybridcattt, line 46: `notShared`? 🤔 😅  a bit confused, could you please explain?

SayTheirNames/Source/Dependency/DependencyContainer.swift
```diff
@@ -24,23 +24,63 @@
...
+    func add<S: Dependency>(handle: S) {
+        let key = String(describing: handle.self)
+        self.factoryDict[key] = handle
+    }
+
+    func resolve<S: Dependency>(_ type: S.Type) -> S {
+        let key = String(reflecting: type)
+        let component = self.factoryDict[key]
+        
+        return component as! S
```
hybridcattt, line 84: guard & fatal error would be better here (minor thing though)

SayTheirNames/Source/Dependency/DependencyContainer.swift
evilpenguin: We can indeed. Only thing is we cannot have a `Dependency` use `@DependencyInject`.

SayTheirNames/Source/Dependency/DependencyContainer.swift
evilpenguin: I mean it is shared but not realllly shared. Haha. Ill change it back.

SayTheirNames/Source/Dependency/DependencyContainer.swift
Kilo-Loco: I think as the amount of dependencies grows, using lazy would be alright so we aren't bringing these services into memory. Overall, I can go either way

SayTheirNames/Source/Dependency/Networking/NetworkRequestor.swift
evilpenguin: Yup; gonna remove all of this logic into another PR for us.

SayTheirNames/Source/Dependency/DependencyContainer.swift
evilpenguin: Added

SayTheirNames/Source/Dependency/DependencyContainer.swift
evilpenguin: If we go lazy we can have a `Dependency`use `@DependencyInject`. Im ok with either our.

SayTheirNames/Source/Dependency/Networking/NetworkRequestor.swift
evilpenguin: All moved.

SayTheirNames/Source/Dependency/DependencyContainer.swift
```diff
@@ -0,0 +1,88 @@
...
+
+// MARK: - DependencyContainer
+final class DependencyContainer: Dependency {
+    let navigator = Navigator()
+    let image = ImageService()
+    let dateFormatter = DateFormatterService()
+    let network = NetworkRequestor()
+    
+    // MARK: - Init
+    init() {
```
biscuitehh, line 40: Would it be possible to pass in Navigator, ImageService, etc. in the initializer instead of instantiating them as properties? It's more of a preference thing, but it makes it a little less bound to those services and adds a little more visibility at the DependencyContainer call site. Also, if things like ImageService become protocols it'll make it way easier to swap in testing components for tests and alternative implementations.

hybridcattt:
oooops didn't notice it's still a draft. merged! 🚀

SayTheirNames/Source/Dependency/DependencyContainer.swift
biscuitehh: FWIW it doesn't have to happen on this pull-request, but just a suggestion for the future :)

PR #114, opened 2020-06-07
[CHORE] Fix fetch donations endpoint request

Kilo-Loco:
- The `DonationType` had an id of type `String` and it needed to be converted to `Int`
- Both `Donation.person` and `Donation.type` need to be optional. It seems as if they aren't guaranteed fields to be returned by the API
- Some smaller fixes in regards to line length and handling the new optional `Person` for `Donation` objects


SayTheirNames/Source/Model/Donation.swift
```diff
@@ -29,11 +29,13 @@ public struct Donation: Decodable {
     let title: String
     let description: String
     let link: String
-    let person: Person
-    let type: DonationType
+    let person: Person?
+    let type: DonationType?
+    let bannerImagePath: String?
```
evilpenguin, line 34: Do we want these are optional?

Kilo-Loco: I dont think we want them as optional, but the current data set is providing null for some objects and essentially breaking the whole screen when there isnt a value for one object

ohwhen: don't see an issue if the api is returning these as null?

PR #117, opened 2020-06-07
Hide location filters for v1

hybridcattt:
In a PM meeting it was decided to not have this on the home screen for v1, as the filtering feature needs more refining to be usable.

Didn't want to delete good code that we will bring back soon. Not keeping it just hidden not to have the collection view trigger delegate, sizing, etc.

UPD: also hid bookmark buttons for the same reason.


hybridcattt:
Simulator Screen Shot - iPhone 11 - 2020-06-07 at 21 11 44

ohwhen:
This seems a rather crude way to do it, would it maybe be a better idea to check the items of the datasource and if there are none, we'll hide the view for now?

hybridcattt:
it is a dirty way indeed. 
Though I'd rather explicitly disable it than build an optional hiding logic just for this - it might be confusing to one reading code why that functionality is there.
let me think of how to disable it better

ohwhen:
@hybridcattt i'm introducing a FeatureFlags in #118 which could be used.

SayTheirNames/Source/Controller/Home/HomeView.swift
```diff
@@ -42,15 +45,16 @@ final class HomeView: UIView {
...
-        locationLayout.scrollDirection = .horizontal
-        locationLayout.sectionInsetReference = .fromContentInset
-        let inset = Theme.Components.edgeMargin
-        locationLayout.sectionInset = UIEdgeInsets(0, inset, 0, inset)
-        let locationCollectionView = UICollectionView(frame: .zero, collectionViewLayout: locationLayout)
-        locationCollectionView.contentInsetAdjustmentBehavior = .always
-        return locationCollectionView
+    lazy private(set) var locationCollectionView: UICollectionView? = {
+        return nil
+//        let locationLayout = UICollectionViewFlowLayout()
```
Kilo-Loco, line 50: Just a thought that we should keep `TODO` next to commented out code so it's easier to know that it's not dead code and when it should be implemented
```suggestion
//        TODO: Add back in when filters are enabled
//        let locationLayout = UICollectionViewFlowLayout()
```

SayTheirNames/Source/Controller/Home/HomeView.swift
```diff
@@ -138,11 +142,11 @@ final class HomeView: UIView {
         return searchButton
     }()
     
-    let separator: UIView! = {
-        let separator = UIView()
-        separator.backgroundColor = UIColor.STN.separator
-        return separator
-    }()
+//    let separator: UIView! = {
```
Kilo-Loco, line 145:
```suggestion
//    TODO: Add in when filters are reimplemented
//    let separator: UIView! = {
```

SayTheirNames/Source/Controller/Home/HomeView.swift
```diff
@@ -179,21 +183,22 @@ final class HomeView: UIView {
...
-            leading: collections.leadingAnchor,
-            trailing: collections.trailingAnchor,
-            size: Theme.Screens.Home.LocationView.size)
-        separator.anchor(
-            superView: collections,
-            top: locationCollectionView.bottomAnchor,
-            leading: collections.leadingAnchor,
-            trailing: collections.trailingAnchor,
-            size: Theme.Screens.Home.SeparatorView.size)
+//        locationCollectionView.anchor(
```
Kilo-Loco, line 186:
```suggestion
//    TODO: Add in when filters are reimplemented
//        locationCollectionView.anchor(
```

ohwhen:
Updating this PR with the changes from #118....

SayTheirNames/Source/Controller/Home/HomeView.swift
hybridcattt: Left a comment on top of the file, might not be very visible. I'll commit this suggestion though anyway 👍

PR #118, opened 2020-06-07
[FEATURE] Hide bookmarks for v1 launch

ohwhen:
This PR introduces a FeatureFlags to be used to enable/disable features within the app. Additionally it uses the FeatureFlags.boomarksEnabled in two places to hide the button to bookmark a person.

Screen Shot 2020-06-07 at 3 34 47 PM

### DIFF
--- a/SayTheirNames/Source/FeatureFlags.swift
+++ b/SayTheirNames/Source/FeatureFlags.swift
@@ -1,0 +1,30 @@
+//
+//  FeatureFlags.swift
+//  SayTheirNames
+//
+//  Copyright (c) 2020 Say Their Names Team (https://github.com/Say-Their-Name)
+//
+//  Permission is hereby granted, free of charge, to any person obtaining a copy
+//  of this software and associated documentation files (the "Software"), to deal
+//  in the Software without restriction, including without limitation the rights
+//  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+//  copies of the Software, and to permit persons to whom the Software is
+//  furnished to do so, subject to the following conditions:
+//
+//  The above copyright notice and this permission notice shall be included in
+//  all copies or substantial portions of the Software.
+//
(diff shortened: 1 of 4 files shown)


SayTheirNames/Source/FeatureFlags.swift
```diff
@@ -0,0 +1,30 @@
...
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+
+import Foundation
+
+struct FeatureFlags {
```
Kilo-Loco, line 27: What if we made this a dependency instead so they could be configured by the backend in the future?

hybridcattt: Love this approach, and yes!
That's exactly how I usually architect apps too. Though here figured to do quick and dirty, so will close my #117

hybridcattt: Not hiding filters yet? Are you planning to do it next?

ohwhen: Yeah good idea, let's discuss it with the backend team 👍  

To keep things moving, let's land this and we can quickly change it once the API is done?

hybridcattt: of course! actually what I suggested was regardless of the API - flags can be local but still be organised in the dependency container to keep things in one place.

PR #121, opened 2020-06-07
[FEAT] Add Korean localization

cdowney102:
Adding Korean strings. I studied Korean for ~6 years. I also had a close Korean friend double check my work. I will gladly maintain this list and add as necessary!

Also, found and fixed a little typo in the word `verified` in the base file.


jongjinchoi:
Nice work. 🙌  I’m late to add korean.

PR #122, opened 2020-06-07
Remove custom navigation views and convert them to native

kutan74:
Simulator Screen Shot - iPhone 11 Pro - 2020-06-07 at 23 30 18

### DIFF
--- a/SayTheirNames/Source/Controller/Donations/DonationsController.swift
+++ b/SayTheirNames/Source/Controller/Donations/DonationsController.swift
@@ -45,7 +45,7 @@ final class DonationsController: UIViewController {
     
     override func viewDidLoad() {
         super.viewDidLoad()
-        navigationItem.title = "DONATIONS"
+        navigationItem.title = Strings.donations
         configure()
     }
     
--- a/SayTheirNames/Source/Controller/Donations/DonationsController.swift
+++ b/SayTheirNames/Source/Controller/Donations/DonationsController.swift
@@ -45,6 +45,7 @@ final class DonationsController: UIViewController {
     
     override func viewDidLoad() {
         super.viewDidLoad()
+        navigationItem.title = "DONATIONS"
         configure()
     }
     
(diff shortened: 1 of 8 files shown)


SayTheirNames/Source/Controller/Donations/DonationsController.swift
```diff
@@ -45,6 +45,7 @@ final class DonationsController: UIViewController {
     
     override func viewDidLoad() {
         super.viewDidLoad()
+        navigationItem.title = "DONATIONS"
```
hybridcattt, line 48: This can use `Strings.donations`

SayTheirNames/Source/Controller/Home/HomeController.swift
```diff
@@ -64,8 +61,8 @@ final class HomeController: UIViewController {
     override func viewDidLoad() {
         super.viewDidLoad()
         view.backgroundColor = UIColor.STN.black
-        navigationController?.navigationBar.isHidden = true
         searchBar.setup(withController: self)
+        navigationItem.title = "SAY THEIR NAMES"
```
hybridcattt, line 65: same here,  there's an L10N key for this

SayTheirNames/Source/Controller/Home/HomeController.swift
```diff
@@ -121,8 +122,8 @@ final class HomeController: UIViewController {
         peopleCollectionView.accessibilityIdentifier = "peopleCollection"
     }
     
-    // MARK: - IBACTIONS
-    @IBAction func searchButtonPressed(_ sender: Any) {
+    // MARK: - Button Actions
+    @objc func searchButtonPressed(_ sender: Any) {
```
hybridcattt, line 126:
```suggestion
    @objc private func searchButtonPressed(_ sender: Any) {
```

ohwhen:
Please correct me if I'm wrong but I think this broke the search input?

kutan74:
You're right. I've added a card for it on Trello. The previous search bar embedded in HomeController but it looks like the design has changed since then. The new search feature should be in a separate view controller. I'm planning to work on it tomorror

ohwhen:
ok, can we revert this merge?

PR #128, opened 2020-06-08
[CHORE] - Add News Cell Title/Image Loading Functionality

biscuitehh:
Since we only get an URL from the API, I found a SDWebImage plugin that utilizes the LinkPresentation framework on iOS to extract rich information from a webpage. Long term it would be cool if the API/link shortener service handled this for us, but this will do in a pinch.

Also, I added a Cocoapod which is why the diff seems huge.


SayTheirNames/Source/Controller/Home/Person/Views/PersonNewsTableViewCell.swift
```diff
@@ -122,13 +124,19 @@ class PersonNewsTableViewCell: UITableViewCell {
         cellType = type
         collectionView.delegate = self
         collectionView.dataSource = self
-        collectionView.register(cellType: cell)
+        
+        switch type {
+        case .medias:
+            collectionView.register(PersonMediaCollectionViewCell.self, forCellWithReuseIdentifier: PersonMediaCollectionViewCell.reuseIdentifier)
```
hybridcattt, line 130: we have an extension for registering and dequeuing called `Reusable`:
```suggestion
                    collectionView.register(cellType:PersonMediaCollectionViewCell.self
```

SayTheirNames/Source/Controller/Home/Person/Views/PersonNewsTableViewCell.swift
```diff
@@ -51,6 +51,8 @@ class PersonNewsTableViewCell: UITableViewCell {
         return "\(Self.self)"
     }
     
+    private var identifier: String = "PersonNewsCollectionViewCell"
```
hybridcattt, line 54: custom identifiers aren't needed if you use methods from `Reusable` protocol which all cells already conform to

SayTheirNames/Source/STNAppDelegate.swift
```diff
@@ -37,6 +39,10 @@ class STNAppDelegate: UIResponder, UIApplicationDelegate {
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
         NSSetUncaughtExceptionHandler(exceptionHandler)
         
+        // Setup SDWebImage Plugins
+        SDImageLoadersManager.shared.addLoader(SDImageLinkLoader.shared)
+        SDWebImageManager.defaultImageLoader = SDImageLoadersManager.shared
```
hybridcattt, line 44: How does it affect regular image loading if this is set as a `defaultImageLoader`?

biscuitehh: @hybridcattt from my testing it didn't seem to affect normal image loading - I don't know a ton about the internals of SDWebImage but images were cached/loaded appropriately during testing. I'm not sure if it's possible with SDWebImage, but I'd feel a lot better just using it for the URLs if possible.

SayTheirNames/Source/Controller/Home/Person/Views/PersonNewsTableViewCell.swift
```diff
@@ -150,8 +158,20 @@ extension PersonNewsTableViewCell: UICollectionViewDelegate, UICollectionViewDat
     
     // Set the data for each cell (color and color name)
     func collectionView(_ collectionView: UICollectionView, cellForItemAt indexPath: IndexPath) -> UICollectionViewCell {
-        let cell = collectionView.dequeueReusableCell(withReuseIdentifier: cellType.identifier, for: indexPath)
-        return cell
+        switch cellType {
+        case .medias:
+            let cell = collectionView.dequeueReusableCell(withReuseIdentifier: PersonMediaCollectionViewCell.reuseIdentifier,
```
Kilo-Loco, line 163: If we use this convenience method, we wont have to force unwrap the cell
```suggestion
            let cell: PersonMediaCollectionViewCell = collectionView.dequeueCell(for: indexPath)
```

SayTheirNames/Source/STNAppDelegate.swift
hybridcattt: completely agree! I think it's worth looking into.

biscuitehh:
Closing this - after talking with @ohwhen I think it's better to do this without a dependency. It shouldn't take long and it helps prevent unnecessary risk messing around with the SDWebImage setup. Thanks everyone for giving feedback/taking a look 🐕 🌮 🚀 !

PR #129, opened 2020-06-08
[CHORE] Feed data into PersonController

kamenibzv:
Waiting on API to be updated before adding news and media


SayTheirNames/Source/Controller/Home/Person/Views/PersonHashtagTableViewCell.swift
```diff
@@ -54,6 +54,7 @@ class PersonHashtagTableViewCell: UITableViewCell {
     }()
     
     private var identifier: String = ""
+    public var person: Person!
```
Kilo-Loco, line 57: I think it would be more efficient to only store the Hashtags as opposed to the whole `Person` object. I also think it would be nice to initialize with an empty array instead of using a `!`
```suggestion
    public var hashtags = [Hashtag]()
```

Sharkesm: @Kilo-Loco Nice catch 💯

PR #133, opened 2020-06-08
[FEAT] Pagination on Home

hybridcattt:
@OnyekachiSamuel and I have been working on adding pagination support to our controllers.
We started with home, and will make the piece reusable so it can be applied to other screens later today.
Putting this up for early feedback, but please mind that this is very much work in progress. 

To-do left in this PR: 
- [x] Make the component generic and reusable
- [x] Add unit tests

Follow up tasks that will be created in Trello
- [x] Make it so it's not needed to `import Alamofire` when using `NetworkRequestor` (done)
- [ ] Align usage of diffable data source across screens
- [ ] Add unit tests for `CollectionViewManager`
- [ ] Add pagination to donations and petitions (work in progress)


SayTheirNames/Source/Controller/Home/HomeController.swift
```diff
@@ -161,4 +154,13 @@ extension HomeController: UICollectionViewDelegateFlowLayout {
             present(navigationController, animated: true, completion: nil)
         }
     }
+    
+    func collectionView(_ collectionView: UICollectionView, willDisplay cell: UICollectionViewCell, forItemAt indexPath: IndexPath) {
+        guard collectionView === peopleCollectionView else { return }
+        guard peopleDataSourceHelper.section(at: indexPath.section) == .main else { return }
+
```
Mackis, line 161:
```
 guard 
            collectionView === peopleCollectionView else,
            peopleDataSourceHelper.section(at: indexPath.section) == .main 
else { return }
```

SayTheirNames/Source/Controller/Home/HomeController.swift
```diff
@@ -161,4 +154,13 @@ extension HomeController: UICollectionViewDelegateFlowLayout {
             present(navigationController, animated: true, completion: nil)
         }
     }
+    
+    func collectionView(_ collectionView: UICollectionView, willDisplay cell: UICollectionViewCell, forItemAt indexPath: IndexPath) {
+        guard collectionView === peopleCollectionView else { return }
+        guard peopleDataSourceHelper.section(at: indexPath.section) == .main else { return }
+        
+        if indexPath.row == collectionView.numberOfItems(inSection: indexPath.section) - 1 {
+            viewModel.loadNextPage()
```
Mackis, line 163: Maybe we could move this to `collectionView:willDisplayCell:forItemAtIndexPath:` to start the loading of the next page as soon as possible?

Mackis:
Looking great 🚀

SayTheirNames/Source/Controller/Home/HomeController.swift
hybridcattt: it's already been like that, so not sure what you mean 🤔  
now renamed the variable so this became outdated

SayTheirNames/Source/Controller/Home/HomeController.swift
hybridcattt: It's a matter of taste I'd say  :)

SayTheirNames/Source/Controller/Home/HomeController.swift
```diff
@@ -24,15 +24,32 @@
...
     @DependencyInject private var network: NetworkRequestor
     
     required init() {
+        
+        weak var wself: HomeController?
+        
+        self.paginator =
+            Paginator<Person, PersonsResponsePage>(pageLoader: { (link: Link?, completion: @escaping (Result<PersonsResponsePage, Error>) -> Void) in
+            guard let sself = wself else { return }
+
```
mahmoudashraf93, line 41: Can we do something about the `wself` and `sself`?

mahmoudashraf93: `self.paginator =
            Paginator<Person, PersonsResponsePage>(pageLoader: { [weak self](link, @escaping (Result<PersonsResponsePage, Error>) -> Void) in
                guard let self = self else {return}` i meant something like this

mahmoudashraf93: You’re right, my bad 😄
Amazing work BTW 🚀

hybridcattt:
Added tests. I found it easiest to test in one function, as there's no easy way to set up initial state since it's private. If anyone got ideas how to split it, happy to hear them :)

SayTheirNames/Source/Controller/Home/HomeController.swift
ohwhen: @hybridcattt would you mind considering using weakSelf and get rid of wself/sself? Ie something like this

```
        weak var weakSelf: HomeController?
        
        self.paginator =
            Paginator<Person, PersonsResponsePage>(pageLoader: { (link: Link?, completion: @escaping (Result<PersonsResponsePage, Error>) -> Void) in
            guard let self = weakSelf else { return }
            
            // call network
            if let link = link {
                self.network.fetchPeopleWithLink(link, completion: completion)
            }
            else {
                self.network.fetchPeople(completion: completion)
            }
        })
        super.init(nibName: nil, bundle: nil)
        
        weakSelf = self
```

hybridcattt: sure that makes sense, done 👌

SayTheirNamesTests/UtilsTests/PaginatorTests.swift
```diff
@@ -0,0 +1,200 @@
...
+        }
+        XCTAssertEqual(sut.state, .notLoading)
+        
+        sut.loadNextPage()
+        XCTAssertEqual(sut.state, .loading)
+        self.waitForExpectations(timeout: 1, handler: nil)
+        XCTAssertEqual(sut.state, .notLoading)
+    }
+    
+    func testLoadFailing() {
```
mahmoudashraf93, line 84: Can the test function name be more descriptive?

hybridcattt: I don't want to push another change as that will dismiss all reviews again, but it's a good point so I'll fix that some other time soon 🙏

PR #134, opened 2020-06-08
[UI] Modify datasource and layout on DonationsMoreDetails controller.

dasautoooo:
### Changes:
- Change donation detail's image source to donation's bannerImagePath.
- Remove donation detail's outcome section.
- Add DonationTypes to determine which image should be displayed on DonationsMoreDetials VC.
- Remove the first donate button.
- Change the sticky donate button title.

### Screenshots:
Simulator Screen Shot2
Simulator Screen Shot1


hybridcattt:
What is the context for removing the section? Is it because it's excluded from v1?

dasautoooo:
> What is the context for removing the section? Is it because it's excluded from v1?

Because there's no data support this section, it is always empty. We can add this back if the data modal needs it.

dasautoooo:
> > What is the context for removing the section? Is it because it's excluded from v1?
> 
> Because there's no data support this section, it is always empty. We can add this back if the data modal needs it.

The API [data](https://saytheirnames.dev/api/donations) has this `outcome` key, but the value for the key is `null` for all data.
The data model for Donation doesn't have `outcome`.
`   let id: Int
    let title: String
    let description: String
    let link: String
    let person: Person?
    let type: DonationType?
    let bannerImagePath: String?
`

dasautoooo:
it's now using `FeatureFlags` to disable outcome section.

SayTheirNames/Source/Controller/Donations/DonationsMoreDetails/DMDCollectionViewDataSource.swift
```diff
@@ -102,8 +103,7 @@ extension DonationsMoreDetailsController: UICollectionViewDataSource {
                 withReuseIdentifier: DMDPhotoSupplementaryView.reuseIdentifier,
                 for: indexPath) as? DMDPhotoSupplementaryView {
                 
-                let person = donation.person
-                person.flatMap { photoView.configure($0) }
+                photoView.configure(withURL: donation.bannerImagePath ?? "")
```
hybridcattt, line 106: if passing `""`, could it be nil instead? nil is standard in swift for "no value"

SayTheirNames/Source/Controller/Donations/DonationsMoreDetails/DonationsMoreDetailsController+Layout.swift
```diff
@@ -88,7 +88,36 @@ extension DonationsMoreDetailsController {
...
+                section.boundarySupplementaryItems = [titleView]
+                
+                return section
+                
+            // Outcome section layout
+            case DonationSectionLayoutKind.outcome.rawValue:
+                // Item
+                let itemSize = NSCollectionLayoutSize(widthDimension: .fractionalWidth(1.0),
+                                                      heightDimension: FeatureFlags.dmdOutcomeSectionEnabled ?
+                                                        .estimated(200.0) : .absolute(0.1))
```
hybridcattt, line 101: I've looked a bit closer at the code and not sure I'm understanding it right. 
The section is there still, just having cells of height 0.1? It would make more sense to not add the section to the collection view altogether.

SayTheirNames/Source/Controller/Donations/DonationsMoreDetails/DMDCollectionViewDataSource.swift
dasautoooo: Great one! Already covered in the lasted commit.

SayTheirNames/Source/Controller/Donations/DonationsMoreDetails/DonationsMoreDetailsController+Layout.swift
dasautoooo: Agreed. It's in the changes.

SayTheirNames/Source/View/ImageWithBlurView.swift
```diff
@@ -66,9 +66,9 @@ class ImageWithBlurView: UIView {
     }
     
     // MARK: - Method
-    public func setup(_ person: Person) {
-        guard person.images.first != nil else { return }
-        frontImageView.populate(withURL: person.images[0].personURL)
-        bgImageView.populate(withURL: person.images[0].personURL)
+    public func setup(withURL urlString: String?) {
```
ohwhen, line 69: small nitpick, label is withURL but you're passing a string...

dasautoooo: I don't think that's a problem, because that's a URL string. Do you mean to URL instead of String, or just to change the function's outside name? i.e. `public func setup(withURLString string: String?)`

ohwhen: @dasautoooo yeah exactly just change the label to `withURLString`...nice one +1

SayTheirNames/Source/Controller/Donations/DonationsMoreDetails/DonationTypes.swift
```diff
@@ -0,0 +1,40 @@
...
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+
+import Foundation
+
+enum DonationTypes {
```
Mackis, line 27:
```
enum DonationTypes: Int {
    case victims, protesters, movement
    
    var id: Int { return self.rawValue }
}
```

SayTheirNames/Source/Controller/Donations/DonationsMoreDetails/DMDCollectionViewDataSource.swift
```diff
@@ -102,8 +103,13 @@ extension DonationsMoreDetailsController: UICollectionViewDataSource {
...
                 for: indexPath) as? DMDPhotoSupplementaryView {
                 
-                let person = donation.person
-                person.flatMap { photoView.configure($0) }
+                switch donation.type?.id {
+                case DonationTypes.movement.id:
+                    photoView.configure(withURLString: donation.bannerImagePath)
+                default:
+                    photoView.configure(withURLString: donation.person?.images.first?.personURL)
+                }
```
bertadevant, line 111: What happens if this image is nil, do we show an empty view? is there a placeholder?

SayTheirNames/Source/Controller/Donations/DonationsMoreDetails/DonationsMoreDetailsController+Layout.swift
```diff
@@ -33,6 +33,35 @@ extension DonationsMoreDetailsController {
         let layout = UICollectionViewCompositionalLayout {
             (sectionIndex: Int,
             _ : NSCollectionLayoutEnvironment) -> NSCollectionLayoutSection in
+            
+            // Later Feature
```
bertadevant, line 37: Is there a Trello card already to implement this? Might be a good idea to leave the TODO with the link so if people see it, they can check the card and implement it.

SayTheirNames/Source/View/ImageWithBlurView.swift
```diff
@@ -66,9 +66,9 @@ class ImageWithBlurView: UIView {
     }
     
     // MARK: - Method
-    public func setup(_ person: Person) {
-        guard person.images.first != nil else { return }
-        frontImageView.populate(withURL: person.images[0].personURL)
-        bgImageView.populate(withURL: person.images[0].personURL)
+    public func setup(withURLString string: String?) {
+        guard let urlString = string else { return }
```
bertadevant, line 70: I think we should have a placeholder, like the fist image for example instead of just leaving it empty.

dasautoooo:
Do NOT merge right now. It needs some changes.

SayTheirNames/Source/Controller/Donations/DonationsMoreDetails/DMDCollectionViewDataSource.swift
dasautoooo: Yes, there's already a placeholder image to the UIImageView.

SayTheirNames/Source/Controller/Donations/DonationsMoreDetails/DonationsMoreDetailsController+Layout.swift
dasautoooo: The feature layout is already implemented, the reason to disable it is because there's no data support this section at this moment, it will yield this section always be empty.

dasautoooo:
> Do NOT merge right now. It needs some changes.

The alarm has been lifted 😇.

SayTheirNames/Source/Controller/Donations/DonationsMoreDetails/DonationTypes.swift
```diff
@@ -24,8 +24,26 @@
...
+//            "id": 2,
+//            "type": "Protesters"
+//        },
+//        {
+//            "id": 3,
+//            "type": "Movement"
+//        }
+//    ]
+//}
+
```
Sharkesm, line 44: @dasautoooo Can you remove these comments if they don't represent anything

SayTheirNames/Source/Controller/Donations/DonationsMoreDetails/DonationTypes.swift
```diff
@@ -24,8 +24,26 @@
...
 enum DonationTypes: Int {
-    case victims, protesters, movement
+    case victims = 1, protesters, movement
     
     var id: Int { return self.rawValue }
 }
```
Sharkesm, line 49: +1

SayTheirNames/Source/Controller/Donations/DonationsMoreDetails/DonationTypes.swift
dasautoooo: I added this intentionally because I wanna make sure other people know why this should start from 1.

PR #135, opened 2020-06-08
[Chore] adding extensions tests UIEdgeInsetsInitTests

kaushikGit8:
In first PR, started adding tests for Extensions, Utils etc. files.
This PR adds tests for **UIEdgeInsets** extensions.


SayTheirNamesTests/ExtensionsTests/UIEdgeInsetsInitTests.swift
```diff
@@ -0,0 +1,77 @@
...
+        XCTAssertEqual(subject.right, 0)
+    }
+    
+    func test_init_left_isEqualToZero() {
+        XCTAssertEqual(subject.left, 0)
+    }
+    
+    func test_medium_top_isEqualToExpectedValue() {
+        subject = .medium
+        XCTAssertEqual(subject.top, 16)
```
Kilo-Loco, line 60: If we're going to be testing these things, shouldn't we be checking if they are equal to the theme value?
```suggestion
        XCTAssertEqual(subject.top, Theme.Components.Padding.medium)
```

kaushikGit8: Thats very precise, thanks I've updated the tests now.

PR #137, opened 2020-06-08
[FIX] Localize dates

hybridcattt:
revamp of #76 

- Date follows date format set on device
- Caching formatters by purpose to achieve consistent formatting across the app. Caching by date format is not applicable since that is different for different locales.


rodericj:
Yeah this makes a lot more sense.

PR #142, opened 2020-06-09
[CHORE][UI]Dark mode

IgnatovPasha:
1. renames Assets.xcasset to Image.xcassets
2. adds Color.xcassets with colors for [Any, Dark] 
3. replaces the use of constants from enum UIColor.STN with generated assets, like UIColor(asset: STNAsset.Color.background)
4. prepares available views for DarkMode
Screenshot 2020-06-09 at 03 47 40


evilpenguin:
Is there a reason we didn't make `UIColor.STN` return a `UIColor(asset:)`?

IgnatovPasha:
Screenshot 2020-06-09 at 04 04 21

IgnatovPasha:
BTW there is no need to use SwiftGen at all in this case for images and colors, two extensions on UIImage and UIColor would be enough

evilpenguin:
I was chatting with @Kilo-Loco. Lets go with this approach:

Colors generated as assets by SwiftGen, then pass them into the dynamic method that will be a static let which is named after what the color is for.

```
static let label: UIColor = dynamic(
    light: STNColor.black,
    dark: STNColor.white
)
```

evilpenguin:
You can seem some examples in `UIFont.STN`

IgnatovPasha:
So there was no need to make colors as assets at all, and my amendments to UIColor.STN that were ready 4 hours ago were quite ok, as I see from UIFont.STN code... 
Screenshot 2020-06-09 at 04 30 02

IgnatovPasha:
I see... then ask somebody to reset this branch to commit (https://github.com/Say-Their-Name/say-their-names-ios/pull/142/commits/42a1de912267be39e058f577c06b381a42a0e1d5)

evilpenguin:
Gonna leave this here. Want to see if we’ll add this to techdebt for v2 and go with your approach.

evilpenguin:
Great work!

PR #150, opened 2020-06-09
[CHORE] Updating the media in person vc

kamenibzv:


Kilo-Loco:
What are we accomplishing with this commit?

kamenibzv:
> What are we accomplishing with this commit?

Updating the media cell in the person screen with the data from the API

PR #152, opened 2020-06-09
[FEAT] Implement Pagination in the Donations Controller

OnyekachiSamuel:
#### What this PR is about
Implement pagination feature in the Donations Screen

Ticket
[Link](https://trello.com/c/hqnKvZx8/117-add-pagination-to-donationscontroller)


SayTheirNames/Source/Controller/Donations/DonationsController.swift
```diff
@@ -63,6 +76,25 @@ final class DonationsController: UIViewController {
             return cell
         }
 
+        donationManager.didSelectItem = { donation in
```
hybridcattt, line 79: Missing weak self, and in couple other closures too (there’s a retain cycle)

OnyekachiSamuel: Okay, thanks, I will implement the change.

SayTheirNames/Source/Controller/Donations/DonationsController.swift
```diff
@@ -34,8 +38,21 @@ final class DonationsController: UIViewController {
     private let ui = DonationsView()
     private var donations: [Donation]?
 
+    private var paginator: Paginator<Donation, DonationsResponsePage>
+    
     required init() {
+        weak var weakself: DonationsController?
```
Kilo-Loco, line 44: IMO I dont think we should be having this type of logic in the initializer like this. Any way we can change the paginator to be a lazily loaded variable instead?

OnyekachiSamuel: The paginator can't be lazily loaded variable because of the closure and for we to have access to self, the initialization of the view controller has to be complete.

hybridcattt: @Kilo-Loco I see it a matter of taste to whether to create variables in init or with a lazy var. This approach is also used in another class already. It doesn't affect maintainability or how it works, so may I suggest we move on with this PR and if you feel it's worth doing the change then we can create a card in trello for making sure all variables are lazy 🙏

kaushikGit8:
```
private lazy var paginator: Paginator<Person, PersonsResponsePage> = {
        return Paginator<Person, PersonsResponsePage>(pageLoader: {
            [weak self] (link: Link?, completion: @escaping (Result<PersonsResponsePage, Error>) -> Void) in
            guard let strongSelf = self else { return }
            guard let link = link  else { return strongSelf.network.fetchPeople(completion: completion) }
            // call network
            strongSelf.network.fetchPeopleWithLink(link, completion: completion)
        })
    }()
```

You might have looked this, but does above approach break something here? just trying to analyse the working of paginator

PR #153, opened 2020-06-09
[CHORE] : Unit tests for Model

npvisual:
This is a first pass to add **unit tests** for the data model.

It is based on testing the main `Model` objects (like responses from person, petitions, and donations requests) against static JSON strings obtained from the API endpoints.

I have not included too many assertions just in case this isn't the way we want to go. New assertions can easily be added later.

Also looking at possibly migrating those static JSON strings to a file in a next version.

_Note :_ the Search API test was disabled since this is out of the scope for the current version. However, since I already had it in my code and it will be useful for v2, I feel like it would be good to include it.

Screen Shot 2020-06-09 at 5 55 05 PM
Screen Shot 2020-06-09 at 5 55 56 PM


npvisual:
### New commit :
The latest commit uses an asset bundle.  This helps us to move the JSON static string out of the test classes and instead rely on actual JSON files to test each one of the main Model elements.

Each test was moved the the ModelTests class and we're now testing each element in a test method.

Note that the search data test is still disabled (`testDecodeSearchData()`), as per previous comment (since not in v1).

Down the road we can add a script (via SPM ?) to run a check on those files and reload the resulting data from the API endpoints directly if needed. This would allow for an easy maintenance and validation of the backend APIs without needing an HTTP connection every single time.

### Updated test results : 

Screen Shot 2020-06-09 at 9 22 44 PM

Screen Shot 2020-06-09 at 9 22 56 PM

PR #154, opened 2020-06-10
[CHORE] fixed PersonCell layout breaking when image size changes

mahmoudashraf93:
- added a height anchor constraint with an 80% multiplier

Task:- https://trello.com/c/3tvebl9J/110-personcell-layout-breaks-due-to-image-sizing

Before 

Simulator Screen Shot - iPhone 11 - 2020-06-10 at 02 10 33

After

Simulator Screen Shot - iPhone 11 - 2020-06-10 at 02 11 57

### DIFF
--- a/SayTheirNames/Source/View/Home/Person/PersonCell.swift
+++ b/SayTheirNames/Source/View/Home/Person/PersonCell.swift
@@ -108,6 +108,9 @@ final class PersonCell: UICollectionViewCell {
         labelsAndButtonContainer.addSubview(nameLabel)
         labelsAndButtonContainer.addSubview(dateOfIncidentLabel)
         labelsAndButtonContainer.addSubview(bookmarkButton)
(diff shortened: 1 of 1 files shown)


mahmoudashraf93:
Oops, now it looks like this when I enabled accessibility large text

Simulator Screen Shot - iPhone 11 - 2020-06-10 at 02 19 44

kutan74:
I think this will fix the issue https://trello.com/c/mOaYUyR2/88-implement-self-resizing-fonts-using-scaledfont

mahmoudashraf93:
> Everything here looks fine but if this is breaking accessibility, I think that we should have that issue fixed first. I think we should strive for the app to be usable by everybody as opposed to strive for it to be as beautiful as possible.
> 
> 
> 
> What do you think?

Indeed, totally agree.

PR #173, opened 2020-06-10
[CHORE]  Unit Tests for the Deep Linking mechanism.

npvisual:
The test cases rely on properly building a set of `DeepLink`s from the `DeepLinkHandler`'s `handle(urls:)` method.
Assertions will validate that :
   - there should only be one `DeepLink` per test case.
   - the `DeepLink` has the appropriate display class (view controller)
   - the `DeepLink` has the correct scheme and host
   - the `DeepLink` has the correct path location (first part of the path components)

Note that we had to slightly modify the `handle(urlContext:)` method of the `DeepLinkHandler` class (inserting a new intermediary method) to be able to set deep link URLs from the unit tests.

Screen Shot 2020-06-10 at 5 57 13 PM

Screen Shot 2020-06-10 at 5 57 25 PM

### DIFF
--- a/SayTheirNames/Source/Dependency/Deep Linking/DeepLinkHandler.swift
+++ b/SayTheirNames/Source/Dependency/Deep Linking/DeepLinkHandler.swift
@@ -36,25 +36,36 @@ final class DeepLinkHandler: Dependency {
     // MARK: - Public
     
     public func handle(urlContext: Set<UIOpenURLContext>) {
-        for context in urlContext {
-             if let deepLink = self.deepLink(matching: context) {
+        self.handle(urls: Set(urlContext.map { $0.url }))
+    }
+    
+    // We use this method as an entry point for unit tests.
+    // So we technically don't care about the resulting `DeepLink` array
+    // except for the unit tests.
+    @discardableResult
+    public func handle(urls: Set<URL>) -> [DeepLink] {
+        
+        var response: [DeepLink] = []
+        for url in urls {
+             if let deepLink = self.deepLink(matching: url) {
                 self.navigator.handle(deepLink: deepLink)
-                return
+                response.append(deepLink)
             }
         }
+        return response
     }
     
     // MARK: - Private
     
-    private func deepLink(matching context: UIOpenURLContext) -> DeepLink? {
-        guard let scheme = context.url.scheme, let host = context.url.host
+    private func deepLink(matching url: URL) -> DeepLink? {
+        guard let scheme = url.scheme, let host = url.host
         else { return nil }
         
         for deepLinkType in self.deepLinkTypes {
             let details = deepLinkType.details
             
             if details.uses(scheme: scheme, host: host) {
-                var components = context.url.pathComponents.filter { $0 != "/" }
+                var components = url.pathComponents.filter { $0 != "/" }
 
                 // We are navigating to home
                 if components.count == 0 {
--- a/SayTheirNamesTests/ServiceTests/DeepLinkTests.swift
+++ b/SayTheirNamesTests/ServiceTests/DeepLinkTests.swift
@@ -40,16 +40,16 @@ class DeepLinkTests: XCTestCase {
     var dlDonationFloydG: URL!
     
     // The scheme and host that should be used.
-    let dlScheme = "stn"
-    let dlHost = "saytheirname.netlify.app"
+    let dlScheme = DeepLinkSupport.Schema.stn
+    let dlHost = DeepLinkSupport.Host.default
     
     // The relative paths for deep link URLs
-    let dlPathAbout = "about"
-    let dlPathPeopleJaoP = "profile/joao-pedro"
-    let dlPathPetitions = "petitions"
-    let dlPathDonations = "donations"
-    let dlPathPetitionScurlock = "petitions/petition-for-james-scurlock"
-    let dlPathDonationFloydG = "donate/donate-to-george-floyd"
+    let dlPathAbout = DeepLinkSupport.Path.about
+    let dlPathPeopleJaoP = DeepLinkSupport.Path.profile + "/joao-pedro"
+    let dlPathPetitions = DeepLinkSupport.Path.petitions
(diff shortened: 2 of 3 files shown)


SayTheirNames/Source/Dependency/Deep Linking/DeepLinkHandler.swift
```diff
@@ -36,25 +36,36 @@ final class DeepLinkHandler: Dependency {
...
-        for context in urlContext {
-             if let deepLink = self.deepLink(matching: context) {
+        self.handle(urls: Set(urlContext.map { $0.url }))
+    }
+    
+    // We use this method as an entry point for unit tests.
+    // So we technically don't care about the resulting `DeepLink` array
+    // except for the unit tests.
+    @discardableResult
+    public func handle(urls: Set<URL>) -> [DeepLink] {
```
evilpenguin, line 46: This should never return more than one DeepLink.

SayTheirNamesTests/ServiceTests/DeepLinkTests.swift
```diff
@@ -0,0 +1,161 @@
...
+    let dlScheme = "stn"
+    let dlHost = "saytheirname.netlify.app"
+    
+    // The relative paths for deep link URLs
+    let dlPathAbout = "about"
+    let dlPathPeopleJaoP = "profile/joao-pedro"
+    let dlPathPetitions = "petitions"
+    let dlPathDonations = "donations"
+    let dlPathPetitionScurlock = "petitions/petition-for-james-scurlock"
+    let dlPathDonationFloydG = "donate/donate-to-george-floyd"
```
evilpenguin, line 52: For everything above this line... Is there a reason we aren't using the `DeepLinkSupport` enum?

SayTheirNamesTests/ServiceTests/DeepLinkTests.swift
```diff
@@ -0,0 +1,161 @@
...
+        dlAbout = try XCTUnwrap(URL(string: dlScheme + "://" + dlHost + "/" + dlPathAbout))
+        dlPeopleJaoP = try XCTUnwrap(URL(string: dlScheme + "://" + dlHost + "/" + dlPathPeopleJaoP))
+        dlPetitions = try XCTUnwrap(URL(string: dlScheme + "://" + dlHost + "/" + dlPathPetitions))
+        dlDonations = try XCTUnwrap(URL(string: dlScheme + "://" + dlHost + "/" + dlPathDonations))
+        dlPetitionScurlock = try XCTUnwrap(URL(string: dlScheme + "://" + dlHost + "/" + dlPathPetitionScurlock))
+        dlDonationFloydG = try XCTUnwrap(URL(string: dlScheme + "://" + dlHost + "/" + dlPathDonationFloydG))
+
+        // We instantiate the dependency container that will be used to build
+        // the various `DeepLink` objects, via `.deepLinkHandler.handle(urls:)`
+        container = DependencyContainer()
```
evilpenguin, line 69: You could just initialize `DeepLinkHandler`.

SayTheirNamesTests/ServiceTests/DeepLinkTests.swift
npvisual: Not public. I can change that, but wasn't sure if it was by design or not.

evilpenguin: Yeah they can be public for unit testing.

SayTheirNames/Source/Dependency/Deep Linking/DeepLinkHandler.swift
npvisual: Agreed. @hybridcattt and I had that conversation earlier. Since the `scene(_:openURLContexts:)` provides a `Set` I wasn't sure if there would ever be a case where multiple `UIOpenURLContext` was relevant.

It's actually not a big deal for the unit tests because we have an assertion checking on the number of elements in the set. So I left the original code (iterating through the Set) as-is since I didn't want to change the initial intent of your code. But we can change that if you'd like.

SayTheirNamesTests/ServiceTests/DeepLinkTests.swift
npvisual: I added the file to the test target. Will change the code to use that and re-submit.

SayTheirNamesTests/ServiceTests/DeepLinkTests.swift
npvisual: Same as above. Not public. But I'll change the permissions and add the file to the test target.

SayTheirNamesTests/ServiceTests/DeepLinkTests.swift
npvisual: Added `DeepLinkSupport` as requested. Good call 🙌 !
See 3183342fb3508adce5a86bb9d442e4ee8ceb48b1

SayTheirNamesTests/ServiceTests/DeepLinkTests.swift
npvisual: This however was a quickly unraveling... Since `DependencyContainer()` is final, I don't need to import anything to use it. However when I start using `DeepLinkHandler` directly, I have to pull **a lot** of files into the test target. So I decided against it considering the cost vs. reward.

SayTheirNames/Source/Dependency/Deep Linking/DeepLinkHandler.swift
npvisual: Note : if we decide to change this, let's do this in another PR so as to clearly point out the reason for the change (otherwise it will get drown in here).

npvisual:
After commit 3183342fb3508adce5a86bb9d442e4ee8ceb48b1 : 

Screen Shot 2020-06-10 at 9 19 35 PM

Screen Shot 2020-06-10 at 9 20 17 PM

PR #176, opened 2020-06-11
[CHORE] Added some missing FR L10n strings.

npvisual:
As per [Trello card](https://trello.com/c/7QXQVOwt) requesting to verify localization for v1, the FR localization file was brought up-to-date with the base L10n file. 

Also :
  - removed the "settings" strings in the FR and pt-BR files since they are not used anymore.
  - added missing translation "viewGoFundMePage" for the button label.


SayTheirNames/Resources/Base.lproj/Localizable.strings
```diff
@@ -41,3 +41,4 @@
 "slack_url" = "https://saytheirnames.slack.com/join/shared_invite/zt-eqjuatz7-fgh3zPRXIKiiXsC1Vf3oZA#/";
 "more_history.title" = "SOME HISTORY ON THE SAY THEIR NAME MOVEMENT";
 "massive_thank_you" = "A MASSIVE THANK YOU TO EVERYONE INVOLVED";
+"viewGoFundMePage" = "VIEW GOFUNDME PAGE";
```
Angel5215, line 44: I think other strings in this file are following a snake_case approach instead of a CamelCase. Maybe we should change it to make it more consistent?

SayTheirNames/Resources/fr.lproj/Localizable.strings
```diff
@@ -24,12 +23,21 @@
...
+"person.news" = "Les Actualités";
 "person.media" = "Média";
 "person.hashtags" = "Hashtags des réseaux sociaux";
+"get_involved_dev.title" = "Participer en tant que développeur web";
+"get_involved_design.title" = "Participer en tant que webdesigner";
+"view_repo" = "VOIR LE CODE SOURCE";
+"join_slack" = "REJOINDRE NOTRE CANAL SUR SLACK";
+"more_history.title" = "L'HISTOIRE DU MOUVEMENT DIS LEURS NOMS";
+"massive_thank_you" = "UN GRAND MERCI A TOUS LES CONTRIBUTEURS";
+"viewGoFundMePage" = "LIEN DE LA CAMPAGNE GOFUNDME";
```
Angel5215, line 43: Same in here.

SayTheirNames/Resources/Base.lproj/Localizable.strings
npvisual: @Angel5215 : will do. Thanks !

PR #177, opened 2020-06-11
[FEAT] Added support for sharing 

ohwhen:
This PR adds a SharingService to the dependency container and injects this service to Persons/Donations/Petitions controllers. 

Each controller passes its own Shareable object to the SharingService, which creates an UIActivityViewController that can be presented.


SayTheirNames/Source/Dependency/ShareService.swift
```diff
@@ -0,0 +1,34 @@
...
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+
+import UIKit
+
+final class ShareService: Dependency {
+    public var items: [Shareable]?
+
```
Sharkesm, line 29: Are we going to use this property?

hybridcattt:
Is it expected that sharing a donation or a petition share a general `/donate` & `/sign` links and not links to the respective item?

ohwhen:
> Is it expected that sharing a donation or a petition share a general `/donate` & `/sign` links and not links to the respective item?

The value what's being shared is coming from the back-end, it seems the value in the Shareable is a generic value. I can ask the back-end team if that is expected.

SayTheirNames/Source/Dependency/ShareService.swift
ohwhen: @Sharkesm yeah I thought about removing it, but I figured it would be a way to pass on a Shareable item between two vc's.

hybridcattt:
> > Is it expected that sharing a donation or a petition share a general `/donate` & `/sign` links and not links to the respective item?
> 
> The value what's being shared is coming from the back-end, it seems the value in the Shareable is a generic value. I can ask the back-end team if that is expected.

would guess not, since on the web it works as expected (they use all other links except for base). hope they fix it soon 🤞

PR #182, opened 2020-06-11
[FIX] Reload lists

hybridcattt:
When there's no data in the lists, try to reload on appear (i.e. switching tabs). Handy when there's been no connection before.


hybridcattt:
fixing the branching....

PR #185, opened 2020-06-11
[CHORE] Finalise the More screen

Sharkesm:
- Organize the code-base structure on the MoreController  
- Add twitter section 
- Link all actions to external links 
- Add a custom tap gesture to propagate a value to `UIButton` actions


SayTheirNames/Resources/Base.lproj/Localizable.strings
```diff
@@ -33,12 +33,24 @@
 "person.news" = "News";
 "person.media" = "Media";
 "person.hashtags" = "Social Media Hashtags";
+
+// Get Involved - Slack
+"get_involved.title" = "HOW DO I GET INVOLVED";
```
hybridcattt, line 38: missing question mark

SayTheirNames/Source/Controller/More/MoreView.swift
```diff
@@ -84,119 +143,76 @@ final class MoreView: UIView {
...
+    func getActionButton(for section: MoreSection) -> UIButton {
+        let button = UIButton()
+        button.translatesAutoresizingMaskIntoConstraints = false
+        button.titleLabel?.font = UIFont.STN.sectionHeader
+        button.layer.borderWidth = 2
+        button.backgroundColor = STNAsset.Color.actionButton.color
+        button.setTitleColor(STNAsset.Color.actionButtonTint.color, for: .normal)
+        button.heightAnchor.constraint(equalToConstant: 50).isActive = true
+        
+        let tapGesture = STNTapGestureRecognizer(target: self, action: #selector(openLink(sender:)))
```
hybridcattt, line 189: what is the reason for using a gesture recogniser and not a button action? Then it can't be tapped with voice control, there's no highlighting, etc.

SayTheirNames/Resources/Base.lproj/Localizable.strings
```diff
@@ -33,12 +33,24 @@
...
+
+// Get Involved - Developer
 "get_involved_dev.title" = "HOW DO I GET INVOLVED AS A DEVELOPER?";
-"get_involved_design.title" = "HOW DO I GET INVOLVED AS A DESIGNER?";
-"view_repo" = "VIEW OUR REPO";
-"repo_url" = "https://github.com/Say-Their-Name/say-their-names-ios";
-"join_slack" = "JOIN OUR SLACK CHANNEL";
-"slack_url" = "https://saytheirnames.slack.com/join/shared_invite/zt-eqjuatz7-fgh3zPRXIKiiXsC1Vf3oZA#/";
+"get_involved_dev.button" = "VIEW OUR REPO";
+"get_involved_dev.url" = "https://github.com/Say-Their-Name/say-their-names-ios";
```
hybridcattt, line 45: shouldn't this link to the org? not only iOS developers use iOS apps 😊 
the text then could be `VIEW OUR GITHUB` ?

SayTheirNames/Resources/Base.lproj/Localizable.strings
```diff
@@ -33,12 +33,24 @@
...
+
+// Get Involved - Twitter
+"get_involved_twitter.title" = "JOIN US ON TWITTER";
+"get_involved_twitter.button" = "FOLLOW US";
+"get_involved_twitter.url" = "https://twitter.com/SayTheirName_io";
+
 "more_history.title" = "SOME HISTORY ON THE SAY THEIR NAME MOVEMENT";
 "massive_thank_you" = "A MASSIVE THANK YOU TO EVERYONE INVOLVED";
+"more_history_desc" = "Say Their Names is a social movement to raise awareness for all the black victims of racial inequality all around the world. This movement is in support of the overall Black Lives Matter movement and aims to highlight individuals affected by anti-Black Violence. Say Their Names movement calls attention and demands justice for all victims.";
```
hybridcattt, line 54: could you add these keys to other languages as well 🙏  so at least they fall back to English and not showing up like "more_history_desc"

SayTheirNames/Source/Controller/More/MoreView.swift
```diff
@@ -84,119 +143,76 @@ final class MoreView: UIView {
...
-    private func actionButton(_ section: section) -> UIButton {
-        let actionButton = UIButton()
-        actionButton.translatesAutoresizingMaskIntoConstraints = false
-
-        actionButton.titleLabel?.font = UIFont.STN.sectionHeader
-        actionButton.layer.borderWidth = 2
-        actionButton.backgroundColor = STNAsset.Color.actionButton.color
-        actionButton.setTitleColor(STNAsset.Color.actionButtonTint.color, for: .normal)
-        actionButton.heightAnchor.constraint(equalToConstant: 50).isActive = true
+    func getActionButton(for section: MoreSection) -> UIButton {
```
hybridcattt, line 180: super minor - can be private, and naming-wise `makeActionButton` makes more sense because it creates a new instance every time.

SayTheirNames/Source/Controller/More/MoreView.swift
```diff
@@ -84,119 +143,76 @@ final class MoreView: UIView {
...
-        }
-        UIApplication.shared.open(repoURL)
+        return button
     }
     
-    @objc private func openSlack(sender: UIButton) {
-        guard let repoURL = URL(string: Strings.slackUrl) else {
-            assertionFailure("slackUrl was not generated")
+    /// Opens external links and accept custom gesture
+    @objc func openLink(sender: STNTapGestureRecognizer) {
```
hybridcattt, line 209: minor - can be private

SayTheirNames/Source/Controller/More/MoreView.swift
```diff
@@ -51,26 +110,26 @@ final class MoreView: UIView {
...
-            [moreCard,
-             historySection(),
-             developerSection(),
-             designerSection(),
-             thankYouLabel])
-        
-        contentView.axis = .vertical
-        contentView.spacing = Theme.Components.Padding.extraLarge
-        contentView.translatesAutoresizingMaskIntoConstraints = false
+        let stack = UIStackView()
```
hybridcattt, line 113: not something introduced in this PR, but the screen needs some spacing below this stack view: 
Simulator Screen Shot - iPhone 11 - 2020-06-11 at 17 16 29

SayTheirNames/Source/Controller/More/MoreView.swift
Sharkesm: I think all these functions are within private extensions; Proceeding to update the function names!

SayTheirNames/Source/Controller/More/MoreView.swift
Sharkesm: Good point 👍

hybridcattt:
It was pointed out that links don't need to be in localisation files, and I believe the fix is coming as a follow-up.

PR #186, opened 2020-06-11
Make hashtags copyable

helloITdepartment:
Made the Hashtag in HashtagView copyable for easier sharing. Added a label informing the user that they can tap and hold a hashtag to copy it to the clipboard. Temporarily changes the text in the hashtagLabel to say "copied" and inverts the colors to make it clear something happened (in case the text is hidden under the user's finger).

Included a gif of the feature in action, and it even works in dark mode!
HashtagCopy

### DIFF
--- a/SayTheirNames/Resources/Base.lproj/Localizable.strings
+++ b/SayTheirNames/Resources/Base.lproj/Localizable.strings
@@ -33,6 +33,7 @@
 "person.news" = "News";
 "person.media" = "Media";
 "person.hashtags" = "Social Media Hashtags";
+"person.hashtagCopyTip" = "Press and hold to copy";
 "get_involved_dev.title" = "HOW DO I GET INVOLVED AS A DEVELOPER?";
 "get_involved_design.title" = "HOW DO I GET INVOLVED AS A DESIGNER?";
 "view_repo" = "VIEW OUR REPO";
--- a/SayTheirNames/Source/View/Hashtag/HashtagTableViewCell.swift
+++ b/SayTheirNames/Source/View/Hashtag/HashtagTableViewCell.swift
@@ -47,6 +47,17 @@ class HashtagTableViewCell: UITableViewCell {
         return label
     }()
     
+    lazy var hashtagCopyTipLabel: UILabel = {
+        let label = UILabel()
+        label.text = L10n.Person.hashtagCopyTip.localizedUppercase
+        label.textColor = UIColor(asset: STNAsset.Color.strongHeader)
+        label.font = UIFont.STN.bannerSubitle
+        label.numberOfLines = 1
+        label.minimumScaleFactor = 0.5
+        label.adjustsFontSizeToFitWidth = true
+        return label
+    }()
+    
     lazy var collectionView: UICollectionView = {
         let layout = UICollectionViewFlowLayout()
         layout.scrollDirection = .horizontal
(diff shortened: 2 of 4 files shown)


SayTheirNames/Source/View/Hashtag/HashtagView.swift
```diff
@@ -68,6 +71,35 @@ class HashtagView: UIView {
...
     }
+    
+    private func setupLongPressGestureRecognizer() {
+        addGestureRecognizer(UILongPressGestureRecognizer(target: self, action: #selector(copyHashtagToClipboard)))
+    }
+    
+    @objc private func copyHashtagToClipboard(gestureRecognizer: UIGestureRecognizer) {
+        
+        if gestureRecognizer.state == UIGestureRecognizer.State.began {
+            print("Copying \(hashtagLabel.text ?? "") to clipboard")
```
Mackis, line 82: Please remove this print statement

SayTheirNames/Source/View/Hashtag/HashtagView.swift
```diff
@@ -68,6 +71,35 @@ class HashtagView: UIView {
...
+            let pasteBoard = UIPasteboard.general
+            if let text = hashtagLabel.text {
+                originalText = text
+                pasteBoard.string = originalText
+                UIView.animate(withDuration: 0.5) {
+                    self.hashtagLabel.text = "Copied"
```
Mackis, line 88: This should be added to localisations

SayTheirNames/Source/View/Hashtag/HashtagView.swift
```diff
@@ -39,10 +39,13 @@ class HashtagView: UIView {
         return label
     }()
 
+    private var originalText: String!
```
Mackis, line 42: This is only used in `copyHashtagToClipboard`. Can we move the declaration into that function?

SayTheirNames/Source/View/Hashtag/HashtagTableViewCell.swift
```diff
@@ -47,6 +47,17 @@ class HashtagTableViewCell: UITableViewCell {
         return label
     }()
     
+    lazy var hashtagCopyTipLabel: UILabel = {
```
Mackis, line 50: Do you mind changing these to `private let` instead of `lazy var`

SayTheirNames/Source/View/Hashtag/HashtagView.swift
helloITdepartment: Done

SayTheirNames/Source/View/Hashtag/HashtagView.swift
helloITdepartment: Done

SayTheirNames/Source/View/Hashtag/HashtagView.swift
helloITdepartment: From what I understand, pressing down on the button and letting go are two separate calls from the gesture recognizer. So if it were only declared inside the function, when the function got called with .state == .ended, it will have been redeclared and won't have anything to put in the label

ohwhen: Can we please get rid of this force unwrap?

SayTheirNames/Source/View/Hashtag/HashtagView.swift
```diff
@@ -39,10 +39,13 @@ class HashtagView: UIView {
         return label
     }()
 
+    private var originalText = "#JUSTICEFORFIRST"
```
evilpenguin, line 42: If this never changes lets make it a constant.

helloITdepartment: It does change- on line 84, when the user taps on a hashtag, it gets changed to the actual text of the hashtag so it can be put back when they let go

SayTheirNames/Resources/Base.lproj/Localizable.strings
```diff
@@ -34,6 +34,8 @@
 "person.news" = "News";
 "person.media" = "Media";
 "person.hashtags" = "Social Media Hashtags";
+"person.hashtagCopyTip" = "Press and hold to copy";
```
npvisual, line 37: This wasn't copied over to the other languages or re-generated by SwiftGen to create the corresponding enum value in L10n. Is that ok ?

PR #192, opened 2020-06-11
[CHORE] L10n finalized migration away from Strings...

npvisual:
… and corrected issues with SwiftGen from base translation.

So : 

- `L10n` instances were renamed again since there was some flow back from correcting the errata in PR#185.
- added the proper keys in the localization base file so that SwiftGen could generate the correct enum used in PR#185.

That was tedious... 🥺


npvisual:
Screen Shot 2020-06-11 at 2 55 32 PM

PR #195, opened 2020-06-11
[CHORE] Remove .red and .yellow colors.

npvisual:
Changes : 
 - Removed the .red and .yellow colors from the asset catalog
 - Ran SwiftGen to remove the colors from the enum
 - Replaced the 1 usage of .red with .cellBackground


SayTheirNames/Resources/Generated/strings.swift
```diff
@@ -91,12 +91,12 @@ internal enum L10n {
     internal static let age = L10n.tr("Localizable", "person.age")
     /// Children
     internal static let children = L10n.tr("Localizable", "person.children")
-    /// Social Media Hashtags
-    internal static let hashtags = L10n.tr("Localizable", "person.hashtags")
-    /// Press and hold to copy
-    internal static let hashtagCopyTip = L10n.tr("Localizable", "person.hashtagCopyTip")
     /// Copied
     internal static let copied = L10n.tr("Localizable", "person.copied")
+    /// Press and hold to copy
```
npvisual, line 96: Note : for some reason SwiftGen is just reordering those properties. Not sure why it like that better, but figured everyone had the same issue (this is a 0-sum game, i.e. no real changes)

PR #202, opened 2020-06-12
[CHORE] Update Deeplink for new host and scheme

evilpenguin:
### Info
- New scheme is `stnio`
 -- Worried about scheme collusion.
- New host is `saytheirnames.io`

### Test
- `xcrun simctl openurl booted "stn://saytheirnames.io/about"`
- `xcrun simctl openurl booted "stnio://saytheirnames.io/about"`


evilpenguin:
> I'd prefer `stn://` but I trust Mr 🐧

I changed it back. We can make a mental note for this.

PR #205, opened 2020-06-12
[UI] Ui/intro header text

FranckNdame:
## Changes
- Added Header to People collection view
- Set navigation title to bold like in Figma designs.
<br />


SayTheirNames/Source/View/Home/Hero/HeroTextCell.swift
```diff
@@ -0,0 +1,63 @@
...
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+
+import UIKit
+
+final class HeroTextCell: UICollectionViewCell {
+    
+    private let title: UILabel = {
+       let label = UILabel()
+        label.font = UIFont(name: "Karla-Bold", size: 21)
```
hybridcattt, line 31: here should use one of the resizable fonts - `UIFont.STN.xxx` (whichever one fits best)

SayTheirNames/Source/View/Home/Hero/HeroTextCell.swift
```diff
@@ -0,0 +1,63 @@
...
+        title.text = data.title
+        body.text = data.body
+    }
+    
+    fileprivate func configureView() {
+        layer.borderWidth = 1.5
+        layer.borderColor = UIColor.black.withAlphaComponent(0.8).cgColor
+        title.anchor(
+            superView: self, top: topAnchor, leading: leadingAnchor,
+            padding: .init(top: 24, left: 16))
```
hybridcattt, line 57: It's better to use the spacings from `Theme` constants

FranckNdame: True!

ohwhen:
Would it be more idiomatic to add these as supplementary views instead of cells?

hybridcattt:
> Would it be more idiomatic to add these as supplementary views instead of cells?

@FranckNdame showed me the code yesterday while it was WIP and we discussed that idea, so I can perhaps answer to unblock this.
This collection has 2 sections, one of them is hidden temporarily. So it can't be made a header to either of the existing sections. That's why it's a new section that's on top; and then whether it's a cell or a supplementary I think doesn't make a big difference.

PR #211, opened 2020-06-12
[LOCALIZATION] Adding Dutch 🇳🇱  as a locale.

ohwhen:
This PR adds Dutch 🇳🇱  as a locale.


SayTheirNames.xcodeproj/project.pbxproj
```diff
@@ -196,6 +196,8 @@
 /* Begin PBXFileReference section */
 		00E6A528018ECB54700AB99B /* Pods-SayTheirNamesTests.release.xcconfig */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.xcconfig; name = "Pods-SayTheirNamesTests.release.xcconfig"; path = "Target Support Files/Pods-SayTheirNamesTests/Pods-SayTheirNamesTests.release.xcconfig"; sourceTree = "<group>"; };
 		1846CA462491F17D00792086 /* ShareService.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = ShareService.swift; sourceTree = "<group>"; };
+		1846CA482493E9BA00792086 /* nl */ = {isa = PBXFileReference; lastKnownFileType = text.plist.strings; name = nl; path = nl.lproj/LaunchScreen.strings; sourceTree = "<group>"; };
```
npvisual, line 199: I don't think we need the LaunchScreen to be l10n'd since there's no text on it.

SayTheirNames/Resources/Base.lproj/Localizable.strings
```diff
@@ -130,3 +130,24 @@
 
 /* loco:5ee2ce86663b704aa41688f2 */
 "sign" = "Sign this petition";
+
```
npvisual, line 133: I assume those were part of a different effort to add new keys ?

hybridcattt: indeed!

PR #215, opened 2020-06-12
[FIX] Disable ko & pt-BR locales

hybridcattt:
Disabling korean and pt-BR localizations - they're both translated less than 50%. Can easily re-enable them again later when they're complete. 

Created cards on trello to add missing translations to these languages.


hybridcattt:
> 👍. I assume you wanted to leave the translation files in there ? They were not removed from the repo (just the project tree).

@npvisual yeah exactly, just to avoid extra friction. hoping they'd be back soon!

Issue #240, opened 2020-07-19
Donation button should be hidden if no donation information exists.


karnett:
Upon selecting a person from the home screen, you will see the donation button show. If there is no donation information, tapping the button will not do anything. Found with Darren Cumberbatch's profile.

What is expected of this work?
We should only show the button if there is an action to take, ie: donation information exists.

How does this improve the project?
Improve user experience by ensuring when they tap there button there is an action to take.

Are there any code samples or examples you can think of, in the current repo or in outside repos?
Within the donation block, there is already code to see if donations exist. We need to move it to apply to adding the view instead, not just when the button is tapped.

Trello:
https://trello.com/c/msA5doRU

Mobo24:
Created a pull request for this card.

hybridcattt:
Addressed by #245

PR #245, opened 2020-08-13
#240

Mobo24:
This change hides the Donation button when person does not have donation details.


SayTheirNames/Source/Controller/Home/Person/PersonController.swift
```diff
@@ -153,9 +160,11 @@ class PersonController: UIViewController {
         super.viewDidLoad()
         view.accessibilityIdentifier = "personView"
         self.network.fetchPersonDetails(with: person.identifier) { [weak self] result in
+            guard let self = self else { return }
             switch result {
             case .success(let response):
-                self?.configure(with: response.person)
+                self.configure(with: response.person)
+                self.hideDonationButton()
```
Angel5215, line 167: Maybe we should also hide the button if we get an error response because the button is also showing in that scenario.

Mobo24: Done

SayTheirNames/Source/Controller/Home/Person/PersonController.swift
```diff
@@ -195,6 +206,18 @@ class PersonController: UIViewController {
...
+    func hideDonationButton() {
+           guard self.person.donations.first != nil else {
+             self.donationButtonContainerView.removeFromSuperview()
+             var viewConstraint: NSLayoutConstraint?
+             viewConstraint = self.tableView.bottomAnchor.constraint(equalTo: self.view.safeAreaLayoutGuide.bottomAnchor)
+             viewConstraint?.isActive = true
+               
+             return self.viewBlur.removeFromSuperview()
+           }
+           self.viewBlur.removeFromSuperview()
```
hybridcattt, line 218: I've noticed a few minor issues in this function, the guard is slightly confusing and there's an unnecessary optional var. Also `self.viewBlur.removeFromSuperview()` feels more at home up in the completion. May I suggest to improve the code like this:

```suggestion
        self.viewBlur.removeFromSuperview()
        guard self.person.donations.first == nil else {
            return
        }
        self.donationButtonContainerView.removeFromSuperview()
        let viewConstraint = self.tableView.bottomAnchor.constraint(equalTo: self.view.safeAreaLayoutGuide.bottomAnchor)
        viewConstraint.isActive = true
```

Mobo24: I want the blur to be removed after the table view constraint has been modified.

PR #249, opened 2020-08-17
[FIX] Search results UI

### LINKED ISSUE #248: UI bug: Text is cut off on search results
### Steps

Visit the search results with increased dynamic type size. 

### Actual result

Text is cut off on top and bottom 

### Expected result

Text should not be cut off on top and bottom 

Screenshot 2020-08-17 at 00 18 50

hybridcattt:
Resolves #248 

- Fix cut off text on larger text sizes
- Hide the image, because it gets pixelated when scaled up

Also included related fixes:
- Slightly increase the minimum row height
- Fix rows remaining selected after tap (was visible only in dark mode)
- Use native styling for separator and search field
- Delete unused search bar subclass

| Before | After |
|:------:|:------:|
|Screenshot 2020-08-17 at 00 18 50|Screenshot 2020-08-17 at 17 08 25|


SayTheirNames/Source/Model/Person.swift
```diff
@@ -109,7 +109,7 @@ struct Person: Decodable {
 
 extension Person: SearchResult {
     var image: UIImage? {
-        STNAsset.Image.userCircle.image
+        nil // STNAsset.Image.userCircle.image
```
Sharkesm, line 112: Is there a reason to have this `nil` here?

hybridcattt: The image asset is low-quality, and it's not necessary now since we have only one search result type - person. If it's nil here, the image won't be shown

PR #251, opened 2020-08-18
[FIX] Making sure names fit on home

### LINKED ISSUE #250: UI bug: text is sometimes cut off on home screen
### Steps

Visit the home tab with increased dynamic type size.

### Actual result

Text is cut off on top and bottom, also horizontally, but it's less of an issue

### Expected result

Text should not be cut off on top and bottom.

Screenshot 2020-08-17 at 16 45 49

hybridcattt:
Resolves #250 

The default text size looks the same, but on larger content sizes two lines are now allowed.

| Before | After |
|:------:|:------:|
|Screenshot 2020-08-17 at 16 45 49|Screenshot 2020-08-18 at 15 48 03|

### DIFF
--- a/SayTheirNames/Source/Controller/Home/HomeView.swift
+++ b/SayTheirNames/Source/Controller/Home/HomeView.swift
@@ -84,16 +84,33 @@ final class HomeView: UIView {
             
             case .main:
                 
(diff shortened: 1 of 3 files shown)


Sharkesm (approved):
LGTM ⭐

PR #252, opened 2020-08-18
[FIX] Empty home on slow connection

hybridcattt:
Currently the banner on the home screen shows together with the loaded list.

Showing it right away, so the screen is not just empty while data loads or if loading fails: 

Screenshot 2020-08-18 at 16 07 19


SayTheirNames/Source/Controller/Home/HomeController.swift
```diff
@@ -71,13 +71,19 @@ final class HomeController: UIViewController {
...
+        homeView.translatesAutoresizingMaskIntoConstraints = false
+        view.addSubview(homeView)
+        homeView.peopleDataSource = peopleDataSourceHelper
+
+        homeView.leadingAnchor.constraint(equalTo: view.leadingAnchor).isActive = true
+        homeView.trailingAnchor.constraint(equalTo: view.trailingAnchor).isActive = true
+        homeView.topAnchor.constraint(equalTo: view.topAnchor).isActive = true
+        homeView.bottomAnchor.constraint(equalTo: view.bottomAnchor).isActive = true
+        
         navigationItem.title = L10n.sayTheirNames.localizedUppercase
```
Sharkesm, line 87: @hybridcattt Can we have this part in a grouped function,  Otherwise looks good to me 😃

SayTheirNames/Source/Controller/Home/HomeController.swift
```diff
@@ -71,13 +71,19 @@ final class HomeController: UIViewController {
...
+        
+        homeView.translatesAutoresizingMaskIntoConstraints = false
+        view.addSubview(homeView)
+        homeView.peopleDataSource = peopleDataSourceHelper
+
+        homeView.leadingAnchor.constraint(equalTo: view.leadingAnchor).isActive = true
+        homeView.trailingAnchor.constraint(equalTo: view.trailingAnchor).isActive = true
+        homeView.topAnchor.constraint(equalTo: view.topAnchor).isActive = true
+        homeView.bottomAnchor.constraint(equalTo: view.bottomAnchor).isActive = true
+
```
Sharkesm, line 86: @hybridcattt Can we have this part in a grouped function,  Otherwise looks good to me 😃

Sharkesm (changes requested):

SayTheirNames/Source/Controller/Home/HomeController.swift
hybridcattt: Should be nicer now 🤞

PR #253, opened 2020-08-19
[FIX] Update slack contribution URL

Sharkesm:


hybridcattt (approved):
Would be also good to update the contribution guide. Not a blocker though :)

PR #261, opened 2020-10-15
update gitignore file

ELezov:
Add new rules for cleaning project


ELezov:
Because your CI can't install pods by itself, i can't  apply new .gitignore for cleaning project

hybridcattt:
Pods are checked in on purpose, even the file you added contains this: `We recommend against adding the Pods directory to your .gitignore.`.
I can recommend editing our .gitignore if you want change anything, instead of replacing the whole file with another version copied from somewhere else.

hybridcattt (changes requested):
See above comment

PR #264, opened 2020-10-18
[CHORE] Remove Korean from localise.biz

hybridcattt:
We reached a limit of 10 languages. I think we can remove Korean because it's not finished and unused, and there's no one that can work on it. Clarification: I already removed it from localise.biz, this is just removing it from the script. 
I'm leaving the .strings files in the repo in case we want to reintroduce the language later - they could be reuploaded to localise.biz.


unnamedd (approved):

Sharkesm (approved):

Sharkesm:
Agree 💯

